PR #9, opened 2022-03-23
Added flow to send DocuSign envelope and save envelope ID in ACQPKG record

jasonburkert-ccpo:
Adds a flow named AT-7081 New Acquisition Package to the ATAT scoped application.
This flow addresses and sends an envelope from a DocuSign template and updates the triggering Acquisition Package record with the generated Envelope ID.

Ticket: AT-7081


juliusfitzhugh-ccpo (approved):
Nice work, the flow was tested and provides the expected behavior.

PR #10, opened 2022-03-24
Add Organization and Contacts tables business rules

juliusfitzhugh-ccpo:
Add basic and advanced business rules for `Organization` and `Contacts` tables.

Ticket: AT-7107


jasonburkert-ccpo (approved):
Reviewed business rule behavior in a running environment.

PR #12, opened 2022-03-29
Fix military address business rules

juliusfitzhugh-ccpo:
Update military address city and state business rules
to include a condition if 'Address Type'  is 'Military' the
business rules will run.
- possibility this got removed during a merge conflict

Ticket: AT-7107


jasonburkert-ccpo (approved):
Thanks for the quick fix

PR #18, opened 2022-04-07
Update COR/ACOR Contact information in Salesforce

juliusfitzhugh-ccpo:
Add a table and flow to update COR/ACOR information in Salesforce.

### Update Salesforce COR/ACOR Contact (table)
UpdateSalesforceContactTable_AT-7102
- `Request Body`:  *String* with a max length of 4,000
- `Sys ID`: *Sys Id (GUID)* as the `id`

### ATAT Email Salesforce COR/ACOR Info (flow)
EmailSalesforceFlow_AT-7102
- send email to Salesforce team
  - Input:  `Request Body` of a  created `Update Salesforce COR/ACOR Contact` record

Ticket: AT-7102


zachclark-ccpo:
Reviewing now.

zachclark-ccpo (approved):

juliusfitzhugh-ccpo:
@zachclark-ccpo from our discussion, generally setting the Display value to `true` for a particular column helps differentiate the records that are displayed in a list. In this case the `Sys ID` is an internal column and I don't think SNOW wants you messing with it or displaying it, so even with `Sys ID` set to `true` it doesn't show up.

PR #19, opened 2022-04-07
Update 'Rank' data field and business rules

juliusfitzhugh-ccpo:
Refactor `Rank` column in the `Contacts` table to better handle the uses cases for the different rank components.

Two approaches were considered for this ticket:
1. Create a new table with the rank components 
2. Continue to use the rank column and parse the label field for the rank components

Approach 1 was picked to create a new Military Rank table, since Approach 2 seemed to make things a bit more complex and was a bit ridged since a script would be needed and could easily break if the format of the labels are not consistent. 

### Pros 
- more flexible
- only one field needed in the contacts to reference rank components
- ranks can be queried based on the different components
- cleaner and easier to maintain (no scripts or business rules needed to update fields in the contacts records) 

### Cons 
- needs a new table and creating of rank component records
- possible to add duplicate records if not careful 
- UI needs to update how rank is handled 
- records are not easily transferred using source control (e.g., PDI to DISA instance)

Ticket: AT-7144


jasonburkert-ccpo (approved):
Reviewed changes in a running environment

PR #25, opened 2022-04-12
SNOW Provisioning Scheduled Job

juliusfitzhugh-ccpo:
Provide a scheduled job to poll the provisioning queue and update SNOW records.
- `getProvisioningJobResults` - a function with logic (see code inside of studio or
in the `sys_script_include` file in https://github.com/dod-ccpo/atat-snow/pull/25/commits/6820be09cdcd4bb1f81b310bac9eba49c556c3ba)
- `Get Provisioning Job Results from AWS` - a *scheduled script execution* that 
calls the function (see file inside of studio)
  - scheduled job is inactive and runs every 30 mins (check `active` to see repeat interval)

ScheduledJobExec_AT-7057

### Additional Context
A script includes function `getProvisioningJobResults` was created to manage the logic
of polling the provisioning queue and updating the different records associate with the
polled messages. This was done so that a global `Scheduled Script Execution` can call the 
function while still having the logic contained in the `ATAT` application source control . 
From my understanding if the logic is placed in the `Scheduled Script Execution` file it will 
be global and not available in source control.

Ideally this same logic can be done in a flow which can be seen if you check out the branch 
[feature/AT-7057-provisioning-results-flow](https://github.com/dod-ccpo/atat-snow/tree/feature/AT-7057-provisioning-results-flow) in the DISA storefront but requires a [plugin](https://developer.servicenow.com/dev.do#!/learn/courses/rome/app_store_learnv2_rest_rome_rest_integrations/app_store_learnv2_rest_rome_data_stream_actions/app_store_learnv2_rest_rome_activating_the_integrationhub_enterprise_pack) 
to be activated. The plugin allows for a [JSON parser step](https://docs.servicenow.com/bundle/paris-servicenow-platform/page/administer/flow-designer/reference/json-parser-step-action-designer.html) in actions.

Ticket: AT-7057

### DIFF
--- a/f600233d1b154d507b782f84604bcb12/update/sys_dictionary_x_g_dis_atat_portfolio_acquisition_package.xml
+++ b/f600233d1b154d507b782f84604bcb12/update/sys_dictionary_x_g_dis_atat_portfolio_acquisition_package.xml
@@ -1,0 +1,76 @@
+<?xml version="1.0" encoding="UTF-8"?><record_update>
+    <sys_dictionary action="INSERT_OR_UPDATE" element="acquisition_package" table="x_g_dis_atat_portfolio">
+        <active>true</active>
+        <array>false</array>
+        <array_denormalized>false</array_denormalized>
+        <attributes/>
+        <audit>false</audit>
+        <calculation><![CDATA[(function calculatedFieldValue(current) {
+
+	// Add your code here
+	return '';  // return the calculated value
+
+})(current);]]></calculation>
+        <choice/>
+        <choice_field/>
+        <choice_table/>
+        <column_label>Acquisition Package</column_label>
+        <comments/>
+        <create_roles/>
+        <default_value/>
+        <defaultsort/>
+        <delete_roles/>
+        <dependent/>
+        <dependent_on_field/>
+        <display>false</display>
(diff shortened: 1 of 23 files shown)


f600233d1b154d507b782f84604bcb12/update/sys_dictionary_x_g_dis_atat_portfolio_acquisition_package.xml
```diff
@@ -0,0 +1,76 @@
...
+    <sys_dictionary action="INSERT_OR_UPDATE" element="acquisition_package" table="x_g_dis_atat_portfolio">
+        <active>true</active>
+        <array>false</array>
+        <array_denormalized>false</array_denormalized>
+        <attributes/>
+        <audit>false</audit>
+        <calculation><![CDATA[(function calculatedFieldValue(current) {
+
+	// Add your code here
+	return '';  // return the calculated value
```
zachclark-ccpo, line 11: I am unsure if this is autogenerated

zachclark-ccpo (approved):
Didn't find any issues in the `getProvisioningJobResults` script, looks good.

See review on the ticket here: https://ccpo.atlassian.net/browse/AT-7057?focusedCommentId=330759

PR #26, opened 2022-04-12
Renaming "Current Contract and Recurring Information" 

robertmccardell-ccpo:
Renaming "Current Contract and Recurring Information"  to "Current Contract Information" 
Removing Recurring Information column


juliusfitzhugh-ccpo (commented):
This looks good. My only concern would be the table name still being `x_g_dis_atat_current_contract_and_recurring_information`. It was quicker to update the current table rather than creating a new table, so I understand and think it is okay for now. 

This is something we may want to take note of in the ticket and come back later to replace this table with a new one to better reflective the name (and make shorter).

rmccardell:
@juliusfitzhugh-ccpo  I agree with you...  I don't like the fact that the canonical table name is still the same as the old one but I think for the sake of expediency we can move forward with it this way and then update either once the sprint is over or if time permits in this sprint.

robertmccardell-ccpo:
@juliusfitzhugh-ccpo  I agree with you... I don't like the fact that the canonical table name is still the same as the old one but I think for the sake of expediency we can move forward with it this way and then update either once the sprint is over or if time permits in this sprint.

juliusfitzhugh-ccpo:
@robertmccardell-ccpo I noticed that the Acquisition previously `Current Contract and Recurring Information` column in the Acquisition package needs to be update to reflect the new `Current Contract Information` name. 
AcqPkgColumn_AT-7227

juliusfitzhugh-ccpo (approved):

juliusfitzhugh-ccpo:
@robertmccardell-ccpo fixed merge conflict.

PR #27, opened 2022-04-12
Adding Period of Performance Table

robertmccardell-ccpo:
Adding Period of Performance Table
Adding Period of Performance Column to Acquisition Package table


juliusfitzhugh-ccpo (approved):
Looks good. Table changes match up with ticket.

PR #28, opened 2022-04-13
Add Flow Create Salesforce Contact to ATAT App

zachclark-ccpo:
## Overview
This adds the flow **Create Salesforce Contact (AT-7137)**
image
### Flow Overview
The Flow will create a new Contact in Salesforce when a new Contact record of the Type Mission Owner, COR, or ACOR is created.
After the record is created in Salesforce, we will update the SNOW record in the Contact table with the RecordID from Salesforce.
Salesforce handles duplicate records based on email, returning the error:
`You're creating a duplicate record. We recommend you use an existing record instead.`
Which is caught and logged by the flow if there is an error.
### Changes made to Contacts Table
Added a new Column, Salesforce ID

Ticket: [AT-7137](https://ccpo.atlassian.net/browse/AT-7137)


juliusfitzhugh-ccpo (approved):
Looks good, things are coming together! The flow ran successfully for the first time and failed during the second call. Kayne also confirmed the flow created a contact in Salesforce.

PR #29, opened 2022-04-13
Table and Column updates

robertmccardell-ccpo:
Updating Period of Performance Table to include "Base And Options" column
Adding Contract Type Table
Updating Sensitive Information Table to include "Section 508 Sufficient" column


juliusfitzhugh-ccpo (changes requested):
I would update the `Base and Options` column to a type of `Name-Value Pairs`, other than that everything looks good.

robertmccardell-ccpo:
> I would update the `Base and Options` column to a type of `Name-Value Pairs`, other than that everything looks good.

Done

zachclark-ccpo (approved):

juliusfitzhugh-ccpo (approved):

PR #30, opened 2022-04-13
Contact Table Updates

robertmccardell-ccpo:
Adding phone extension column to support phone extensions
Adding manually entered (true/false) column to support COR/ACOR when the user enters the contact info as opposed to selecting an existing one


zachclark-ccpo:
Need to update the checksum file, locally you need to do:
`git pull origin develop`
And fix the merge conflict by accepting the incoming change.
After that, commit the changes.

zachclark-ccpo (approved):

PR #35, opened 2022-04-19
Reference Contract Type in Acquisition Package

juliusfitzhugh-ccpo:
Reference contract type in an acquisition package.


jasonburkert-ccpo (approved):
Reviewed change in running environment.  Discussed with @robertmccardell-ccpo to be sure this was the right path.

PR #39, opened 2022-05-03
added field to sensitive information table

DGabriel-CCPO:
Changes made for https://ccpo.atlassian.net/browse/AT-7377?atlOrigin=eyJpIjoiZjUxYTJlMjFiYjNiNGMzZDkwMzI1N2IwZmI3YmRiNjIiLCJwIjoiaiJ9


f600233d1b154d507b782f84604bcb12/dictionary/x_g_dis_atat_sensitive_information.xml
```diff
@@ -1,5 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?><database>
     <element label="Sensitive Information" max_length="40" name="x_g_dis_atat_sensitive_information" type="collection">
+        <element label="508AccessibilityRequirements" max_length="1000" name="accessibilityrequirements" type="string_full_utf8"/>
```
juliusfitzhugh-ccpo, line 3: Is there a reason the field type needs to be `String (Full UTF-8)`? If not, I'd suggest sticking with a regular `String` type like the other fields. 

Did this replace the previous `Section 508 Sufficient` field or should both that field and the `508AccessibilityRequirements` be present? The `Section 508 Sufficient` is being used by a flow downstream when creating one of the documents and this would cause that information to no longer being present on the document.

FieldType_AT-7377
Screen Shot 2022-05-03 at 12 48 36 PM
FlowInformation_AT-7377


Was the naming for the *Column Name* done that way intentionally? If not, I would also suggest naming the *Column label* as `Accessibility Requirements 508` (numbers are at the beginning are not captured in the name) so the column name ends up being something like `accessibility_requirements_508` for readability and clarity when trying to access that column in a record. You may have to delete the column and make a new one. 
NewColumn_AT-7377

juliusfitzhugh-ccpo (changes requested):
I left some comments regarding the addition of the new column to make sure I'm understanding if some things were intentional or not.

Also, you will get a `checksum.txt` conflict for each PR and that is normal, but any other conflicts seen might suggest  some unintended changes were made or someone else changed similar files. 

If you have any questions let me know.

f600233d1b154d507b782f84604bcb12/dictionary/x_g_dis_atat_sensitive_information.xml
DGabriel-CCPO: This field does not replace the `Section 508 Sufficient` field. If false is selected, it will be the next input on the following page. i have also deleted that column and created a new one with easier readability.

juliusfitzhugh-ccpo:
See PR #40

PR #41, opened 2022-05-04
Adding new tables to support Static UI data

robertmccardell-ccpo:
Adding List Items, States, and Countries tables to support static data needed for UI dropdowns and selection items


juliusfitzhugh-ccpo:
@robertmccardell-ccpo that is understandable. The free text input is only static data being used by the UI right?

Regarding maintainability each list item would be a record, so it would probably be best to create them in the DISA instance or exported/imported somehow (haven't exported/imported records).  Assuming there isn't much change once creation of the records for the `List items` there shouldn't be much updating or adding (similar to the `Military Rank` table). There is the possibility of duplication, so it is a good idea to filter/search and be cautious before adding any additional records after the initial creation.

Is there anything in particular you were concerned with or noticed regarding maintainability?

robertmccardell-ccpo:
@juliusfitzhugh-ccpo yes it's only static data, shouldn't change frequently or at all unless the countries list for some reason...

juliusfitzhugh-ccpo (changes requested):
Going with the suggestion I provided, create a new table called `List Item` with the following columns:
- `List Type` - Choice
- `Label` - String (max length if needed)
- `Value` - String (max length that is suitable for the free form text, maybe 1,000?)

I would create the table, add some records in your PDI instance, and then test it out with how you are expecting to use them in the UI to make sure it works.

jasonburkert-ccpo:
@robertmccardell-ccpo Had you considered using the `core_country` table?

robertmccardell-ccpo:
@jasonburkert-ccpo  I haven't but I'll take a look...  As long as if fits within the constraints that we have been given for the UI

robertmccardell-ccpo:
@juliusfitzhugh-ccpo @jasonburkert-ccpo   Thanks for the suggestion Jason the country table looks like it'll get the job done! 
image

juliusfitzhugh-ccpo:
A solution was found for using tables already provided by SNOW. Closing PR.

PR #44, opened 2022-05-06
Create Funding Plan Table

robertmccardell-ccpo:


f600233d1b154d507b782f84604bcb12/dictionary/x_g_dis_atat_funding_plan.xml
```diff
@@ -0,0 +1,11 @@
+<?xml version="1.0" encoding="UTF-8"?><database>
+    <element label=" Funding Plan" max_length="40" name="x_g_dis_atat_funding_plan" type="collection">
+        <element label="Acquisition Package" max_length="32" name="acquisition_package" reference="x_g_dis_atat_acquisition_package" type="reference"/>
```
juliusfitzhugh-ccpo, line 3: Nice, catch on the `Acquisition Package` reference. I think it should be the reverse with a new column in the `Acquisition Package` table that references the `Funding Plan` table according to AT-7375. That also makes it easier to reference the data from the acquisition package.
- Update `Acquisition Package` table
  - new column `Funding Plan` - Reference to `Funding Plan` table

juliusfitzhugh-ccpo (changes requested):
Looks good, only a small change.

f600233d1b154d507b782f84604bcb12/dictionary/x_g_dis_atat_funding_plan.xml
robertmccardell-ccpo: @juliusfitzhugh-ccpo  Yeah. So actually I'm slightly confused about how this will work...   There can be many Funding plan tables with attachments to one Acquisition package table...    So will that functionality still be captured by adding a Reference column on the Acquisition Package table?

robertmccardell-ccpo: @juliusfitzhugh-ccpo  Makes total sense.   So I'll just add the list column to the acquisition package table and remove the acquisition package reference in the Funding Plan table.

juliusfitzhugh-ccpo: @robertmccardell-ccpo took a brief look and the changes and they look good. I wasn't sure if you were done, but I would suggest adding the `Funding Plan` table as the reference for the column.
FundingPlan_AT-7375

juliusfitzhugh-ccpo (approved):

PR #48, opened 2022-05-12
Add tables for Description of Work document

jasonburkert-ccpo:
Adds the following tables
- `Classification Level`
- `Environment Instance`
- `Current Environment`
- `Award History`
- `Service Offering`
- `Required Services`

Adds the following references
- `Acquisition Package.Current Environment` > `Current Environment`
- `Acquisition Package.Contract Award` > `Award History`
- `Acquisition Package.Contract Modifications` > `Award History`
- `Acquisition Package.Required Services` > `Required Services`
- `Current Environment.Environment Instances` > `Environment Instance`
- `Environment Instance.Classification Level` > `Classification Level`
- `Required Services.Select Service Offerings` > `Service Offering`
- `Required Services.Applicable Classification Levels` > `Classification Level`
- `Required Services.Applicable Periods` > `Period`

Ticket: AT-7340


jasonburkert-ccpo:
DOW section _3. Background/Current Environment_ data is stored in table `Current Environment` and `Environment Instance`
image

image

jasonburkert-ccpo:
DOW section _4. Performance Requirements_ data is stored in table `Required Services`.
image

This table has three important references (all Lists) so that a group of selected services can be applied to specific classification levels and applicable contract base and option periods:

1. `Service Offering` - this is reference data codifying groups/types of offerings listed in DOW subtasks such as 4.2.2
2. `Classification Level` - this is reference data codifying valid combinations of classification and impact level
3. `Period` - entries define base and options periods, and are referenced in `Period of Performance` for a given package

Business rules will need to enforce that base and option periods listed here only include those referenced in `Period of Performance`.

image


Several rows in table `Required Services` may be required per package to satisfy the flexible subtasks in DOW section 4, therefore `Acquisition Package` contains a List of references to `Required Services`.

I believe this structure will satisfy the majority of 4.2.x

Subtasks 4.2.1 and 4.2.7 contain an additional complication of instances and I was advised that we are still awaiting finalized requirements for them and to concentrate effort elsewhere.

jasonburkert-ccpo:
The data for the table directly under the document title of the DOW is stored in table `Award History`.
image

zachclark-ccpo:
Reviewing now.

zachclark-ccpo:
On the Environment Instance table it doesn't look like **CSP Region** has any options.
image
This is the same behavior for: 
**Performance Tier**
**Storage Type**
Which are fields with the **Choice** data type that have no options to choose.

zachclark-ccpo:
> The data for the table directly under the document title of the DOW is stored in table `Award History`. image

image
Since you can choose 
**Initial Award**
or
**Modification**

That means you can have multiple awards/modifications on the acquisition package right? 

image
Looking at the acquisitions table it appears that the **Contract Award** Reference only lets you select a single award.

image
So we probably might need to update the implementation of the **Contract Award** reference to allow for the full Award history to be displayed on an Acquisition Package

juliusfitzhugh-ccpo (commented):
Nice work, this was almost like a puzzle to piece together everything for the DoW. I did not try to review the XML files. Looking at the DoW document and excel spreadsheet seems like most of the things are covered or will soon be covered. 

I noticed that the `Classification Level` table had a choice field for both the `Classification` and the `Impact Level`. Could this have been grouped into one and have a `Choice` field with a combination of the `Classification` and `Impact level` and then the required columns use those choices? Not saying this needs to be done, just asking.

You mentioned the surge capability is missing in the ticket, I believe that is in the `Requirements Cost Estimate` table.

Are we doing `UPPER_CASE_SNAKE_CASE` for choice fields `Values`? If not then you can ignore this, I just noticed they were in lower `snake_case`.

jasonburkert-ccpo:
Good eye.  The intent is that `Contract Award` would reference a single entry of type _Initial Award_, and `Contract Modifications` would be a list of entries of type _Modification_.

> > The data for the table directly under the document title of the DOW is stored in table `Award History`. image
> 
> image Since you can choose **Initial Award** or **Modification**
> 
> That means you can have multiple awards/modifications on the acquisition package right?
> 
> image Looking at the acquisitions table it appears that the **Contract Award** Reference only lets you select a single award.
> 
> image So we probably might need to update the implementation of the **Contract Award** reference to allow for the full Award history to be displayed (retrievable) on an Acquisition Package

jasonburkert-ccpo:
The specific options for performance tier and storage type hadn't been finalized yet, so I left them empty.  We could imagine some options for csp region, but they'd be tied to a specific CSP.  Basically I added placeholders for what's currently represented in section 4.2.1, but expect it may change.  I mentioned something tangential in slack about storage options and it became clear we just didn't know what the options were yet.

In the end , for CSP region we may need to do something requiring reference data, or maybe just make CSP region a text field.

> On the Environment Instance table it doesn't look like **CSP Region** has any options. image This is the same behavior for: **Performance Tier** **Storage Type** Which are fields with the **Choice** data type that have no options to choose.

jasonburkert-ccpo:
> Nice work, this was almost like a puzzle to piece together everything for the DoW. I did not try to review the XML files. Looking at the DoW document and excel spreadsheet seems like most of the things are covered or will soon be covered.
>
Thanks.  Parts were tricky for sure.
 
> I noticed that the `Classification Level` table had a choice field for both the `Classification` and the `Impact Level`. Could this have been grouped into one and have a `Choice` field with a combination of the `Classification` and `Impact level` and then the required columns use those choices? Not saying this needs to be done, just asking.

Yeah, it could be done that way.  I do like that it is it's own entity that can be referenced by several tables.  Already referenced in two places as I noted.

> 
> You mentioned the surge capability is missing in the ticket, I believe that is in the `Requirements Cost Estimate` table.
>

I was not aware of that.  Will check it out.


> Are we doing `UPPER_CASE_SNAKE_CASE` for choice fields `Values`? If not then you can ignore this, I just noticed they were in lower `snake_case`.

If we are I had forgotten.  I made them lower snake case.

jasonburkert-ccpo:
> Good eye. The intent is that `Contract Award` would reference a single entry of type _Initial Award_, and `Contract Modifications` would be a list of entries of type _Modification_.
> 

I chose to do it this way because of notes Kyle gave about Base Period and Option Periods and keeping them separate.  Seemed  like a similar case.

zachclark-ccpo (approved):
Thank you for addressing my feedback, everything else looked fine to me.

juliusfitzhugh-ccpo (approved):
Thanks for addressing my comments. No of the them were blocking, but wanted to mentioned. Approving to move forward.

jasonburkert-ccpo:
I will update Choice values to UPPER_SNAKE_CASE for consistency.  Thanks @juliusfitzhugh-ccpo

PR #49, opened 2022-05-13
Remove Choice Set and all related options that used to support Rank in the Contacts table

jasonburkert-ccpo:
`Contacts` table used to have column `Rank` of type Choice.  This was replaced by table `Military Rank` in #19.
This removes the entire Choice Set associated with that table and all related choice values.
(where Table=x_g_dis_atat_contacts and Element=rank)


juliusfitzhugh-ccpo (approved):
Thanks for cleaning these up. Now that we haven XML file for the ranks records backed up, we shouldn't need these anymore. Glad we had them up until this point though.

PR #50, opened 2022-05-16
Changed field types from True/False to Choice 

ericyoungquist-ccpo:
Update below columns from `True/False` type to `Choice` type with Yes/No/Unselected options. The choices from `Can Access Package` is being reused for the other column choices. 

### Tables and Columns
Contacts
- `.can_access_package`

Sensitive Information
- `.section_508_sufficient`
- `.pii_present`
- `.baa_required`
- `.potential_to_be_harmful`

GFE Overview
- `.gfe_gfp_furnished`
- `.property_accountable`

PoP
- `.pop_start_request`
- `.recurring_requirement`

Ticket: AT-7435


juliusfitzhugh-ccpo (dismissed):
@ericyoungquist-ccpo I'm not sure if this is just something showing up on my PDI (can't check in `develop`), but it looks like some of the field types still appear as `True/False` field types despite the XML showing them are `Choice` fields (see screenshots below).
#### Tables and Columns
- Contacts
  - `.can_access_package` - still shows `True/False`
- Sensitive Information
    - `.section_508_sufficient` - Good (shows as `Choice`)
    - `.pii_present` - still shows `True/False`
    - `.baa_required` - still shows `True/False`
    - `.potential_to_be_harmful` - still shows `True/False`
- GFE Overview
    - `.gfe_gfp_furnished` - still shows `True/False`
    - `.property_accountable` -  still shows `True/False`
- PoP
    - `.pop_start_request` - Good (shows as `Choice`)
    - `.recurring_requirement` - Good (shows as `Choice`)

StillTrueFalse2_AT-7447
StillTrueFalse_AT-7447


I decided to try and send requests to one of the tables to test how it behaves and it seems to still return only `true` or `false` (see screenshot below). The form also only contains a checkbox and not a selection field. There is a [community post](https://community.servicenow.com/community?id=community_question&sys_id=58da70cbdb919f004fc2f4621f961918) that is old, but suggests deleting the column since it can't be converted from a `True/False` and might cause other issues.
RequestYes_AT-7447
ResponseFalse2_AT-7447

I think each `Column` that you changed that are still of `True/False` type have to be deleted then added back as a `Choice` type. The list of columns affected is above (assuming this is not something only on my PDI). Suggested method to update if needed:
1. Copy `Column` name
2. Delete `Column`
3. Recreate `Column` with exact name it had previously
4. Ensure Choice fields are still there (they should still be present)

**Note:** Using the exact name also helps prevent any name changes that are being used by flows later in the process.
TrueFalseType_AT-7447
ChoiceTypeUpdated_AT-7447

Update 19 May 2022: There was some wonkiness going on with ServiceNow that caused inconsistency across PDI and DISA source control syncing for the `True/False` field types being changed to `Choice` field types when the column uses the same name. The issue was resolved by deleting the application files related to the changed columns (on DISA instance), then those changes where stashed which it seems causes the commits from the active branch to be applied to the DISA instance during the stashing process. Still not completely clear, but issues resolved.

ericyoungquist-ccpo:
@juliusfitzhugh-ccpo all good suggestions, and that's actually what I did for each of the fields. Strange... maybe even with them being deleted and recreated with the *exact same* column name isn't enough - maybe that name still exists deep in the bowels of servicenow somewhere perhaps. They all show properly on my PDI. We can look together tomorrow. Thanks!

juliusfitzhugh-ccpo:
@ericyoungquist-ccpo that's fair it did seem strange when I was looking at things, so not sure if it's some side affect from using a PDI either. Sound good, we can look into it tomorrow.

juliusfitzhugh-ccpo:
The issue mentioned above regarding the `True/False` still showing in the DISA instance or other PDIs aside from the PDI the changes were made in was resolved in a way that is not exactly intuitive. The cause appears to stem from changing a `True/False` to `Choice` type wherein the same name is used. 

### Resolution
Basic overview was deleing application files in the DISA instance and then stashing those changes. This caused the source control committed changes to be applied to the DISA instance for the deleted application files (which were the updated choices). This was stumbled upon and was kind of luck, and a true underlying of what is happening is unclear.

Steps:
- In the DISA instance on the desired branch (the branch for this PR) navigate to application files by using `sys_metadata.list` in the search bar 
- Search Application with`*atat` then filter further by using name of column in Update Name (e.g., `*can_access_package`)
- Delete (twice) all of the files related to the column that had been changed from `True/False` to `Choice`
- Stash changes that appear on the branch
- The files from the source control should be present as `Choice` types.

Delete1_AT-7447
Delete2_AT-7447
SwitchBranch_AT-7447
UpdatedChoice_AT-7447

PR #51, opened 2022-05-17
Add ATAT attachment CORS rule

juliusfitzhugh-ccpo:
Supports uploading user stories in epic AT-7235

Ticket: AT-7465


kylelaker-ccpo (changes requested):
This is not an especially safe change. It allows any application running on `localhost:8080` to make requests to SNOW. I mean, that'd have to be authenticated. But this isn't going to solve problems in any actual environment, right? This just makes it so that the SPA running on a developer's machine can upload attachments to a real ServiceNow? That's dangerous and I don't like it.

If I am misunderstanding, then please correct me. But as I am reading this, I don't think this should be done.

jasonburkert-ccpo:
I've scheduled a meeting with relevant parties to discuss tomorrow.

jasonburkert-ccpo:
During our discussion we decided that this rule shouldn't be applied in the ATAT application scope.  Closing this PR and the ticket.

PR #52, opened 2022-05-17
Add reference to classification level in acquisition package table

DGabriel-CCPO:
`Acquisition Package` table
- `Classification Level` - Reference type

Ticket: AT-7438


juliusfitzhugh-ccpo (approved):
Looks good.

PR #63, opened 2022-05-19
Add exported data for reference tables

jasonburkert-ccpo:
Adds exported data for the following reference tables
- `Classification Level`
- `Military Rank`
- `Service Offering`

Ticket: AT-7486


juliusfitzhugh-ccpo (approved):
Thanks, now we have one location for the important files to quickly access. Might want to make sure that the records we do include here are not sensitive. I think these are fine, but just thinking we should be mindful of the things added here.

jasonburkert-ccpo:
@juliusfitzhugh-ccpo Absolutely.  This isn't the solution for all tables and content.  Good advice to be mindful.

PR #64, opened 2022-05-20
Update column in Current Contract Info table

juliusfitzhugh-ccpo:
Convert `Current Contract Exists` column in the `Current Contract 
Information` table from `True/False` to a `Choice` field type. 

**Note:** Changed DISA instance and ensured it is changed to a `Choice` field. Starting to think auto recovery of `sys_metadata` might be the cause for columns remaining as `True/False` and not in sync with the PDI commit changes (see [auto recovery](https://docs.servicenow.com/bundle/rome-application-development/page/build/applications/reference/r_AutoRecoveryProperties.html)).
CurrentContractExistsChoice


jasonburkert-ccpo (approved):
👍
image

PR #65, opened 2022-05-23
Improve service offerings and groups

jasonburkert-ccpo:
CSPs offer services.  These are categorized in groups in the Description of Work (DOW) document.  Groups are represented by SNOW Choice set `service_offering_group`.  Individual service offerings belonging to those groups are listed in table `Service Offering`.  This PR includes the following improvements to this structure:

We'd like to present `service_offering_group`s in a specific order.
- Sets `sequence` value for choice set `service_offering_group`

We'd like to present individual service offerings within each of those groups in a specific order.
- Adds column `Service Offering.Sequence`

We'd like to display subtext or annotations below names of service offerings
- Adds column `Service Offering.Description`

PR #66, opened 2022-05-23
update export of Service Offering table

jasonburkert-ccpo:
contains values for new sequence column added by #65


jasonburkert-ccpo:
Table contents has changed this today and the export file be updated after merge of #65

PR #67, opened 2022-05-25
Add ATAT Extract DoW Data subflow

juliusfitzhugh-ccpo:
Create `ATAT Extract DescriptionOfWork Data` subflow to send as a payload for the `POST /generate-document` request.
- Add action `Gather DoW data` to use in subflow
- Add util functions (in scoped app under **Server Development** > **Script Includes** in Studio)
  - `getListRecords` - a wrapper for `GlideQuery` to get specified fields from a specific table using `sys_id`
  - `getNestedServiceInfo` - gather nested data for a service record for fields have a `List`
    - `getClassificationInstanceInfo` - populates classification levels and periods
  - `mapPackagePeriods` - map of all periods for  package (can probably utilize more)
  - `removeSysId` - remove `sys_id` from object attached when using `GlideQuery` api
  - `convertToBoolean` - convert `Yes/No/Unselected` or  string `true/false` into booleans otherwise return `null`
 
**Inputs:**
- acquisition package - record

**Output:**
- dow data payload - stringified JSON object

Includes changes made in PRs #70, https://github.com/dod-ccpo/atat-web-api/pull/893, and https://github.com/dod-ccpo/atat-web-api/pull/894.

DowExtractSubflow_AT-7394
DowGatherActionCompose_AT-7394

### Notes: 
Data wrangling inside SNOW was not as simple as I thought. There is definitely room  for improvements with gathering the data in DoW (or future docs). I created two separate scripts to try and help simplify the logic, but could probably be broken up more. 

The `sn_ih.JSONStreamingBuilder().build()` seemed like a reasonable way to structure objects, but things get a little tricky when dealing with arrays. For example, forming an array from an already formatted array with objects without recreating is difficult. Also, the builder seems a bit excessive for creating large JSON objects.

The `Array.Object` output type fo the script step was a bit weird and causing an error, but not sure if the output needs to be structured differently or use something like the `JSONStreamingBuilder()`, but stringified JSON objects was used for simplicity and a work around. 

**Expected Errors:**
- @zachclark-ccpo discovered if no `Classification Level` is provided in the `Environment Instance` record an error occurs. An error is output in the system logs as suggested by @kylelaker-ccpo (e.g., `An environment instance has no classification level and one should be provided. Environment Instance Name: Another Test instance. Sys ID: ec5d313f1be34110c385ea85604bcb0b.`)


f600233d1b154d507b782f84604bcb12/dictionary/x_g_dis_atat_acquisition_package.xml
```diff
@@ -26,36 +27,39 @@
...
         </index>
         <index name="index10">
-            <element name="contract_considerations"/>
+            <element name="requirements_cost_estimate"/>
         </index>
         <index name="index11">
+            <element name="organization"/>
+        </index>
+        <index name="index12">
             <element name="current_environment"/>
```
zachclark-ccpo, line 63: Is this intended behavior?

zachclark-ccpo: Nevermind, it seems index change is related to adding that new table in [d7c7d15](https://github.com/dod-ccpo/atat-snow/pull/67/commits/d7c7d150142a36fbfc6fd39b8567fedd8207d33d)

zachclark-ccpo:
I am not sure this is the intended mapping for `required_training_services` (taken from **CompleteDowPayload.json** linked on ticket)
```
       "required_training_services": {
            "name": "required training",
            "value": "{\"1\":\"basic training\",\"2\":\"intermediate training\",\"3\":\"advanced traning\"}"
        }
```

zachclark-ccpo:
image

Yes this is completed by KO at TO award, I think the mapping doc was edited since I made the spec so some fields might be slightly off, but we can edit the schema to not pass the `task_order_number`.

zachclark-ccpo (approved):

PR #69, opened 2022-05-26
Updating the data model to retool the tables.  We realized that we we…

robertmccardell-ccpo:
Updating the data model to retool the tables.  We realized that we weren't accurately representing the data structure needed to fully capture the UI flow for Required Services...  In the previous model the Required Services table included  Applicable Classification Levels, Applicable Periods,  Selected Service Offerings, Need for entire TO duration,  Other Service Offerings,  and Usage Description.   Upon closer consideration  this structure needs to be updated due to the facts that based on the UI comps,  along with each classification level entry, we require the following items to be specified (Need for entire TO duration, Other Service Offering, Usage Description and Applicable Periods). In this PR the proposal has been made to create a separate table, "Service Classification Instance" that will encapsulate the aforementioned fields.   There will be a reference from the Required Services table  to a list of Service Classification Instances.   The Required Services table will lose the following columns (Need for entire TO duration,  Other Service Offerings,  and Usage Description) which have been moved over to the Service Instance Classification Instance table.   
image


zachclark-ccpo:
Can this PR be closed? @robertmccardell-ccpo @jasonburkert-ccpo

jasonburkert-ccpo:
Replaced with #70

PR #77, opened 2022-06-09
AT-7452: Define Costs Table Schema 

jeffsegal-ccpo:
Added a new Table name Costs which will contain the cost data we receive from the CSPs. Also added a Computed Name column to Organizations & made it the Display column for the Table so that we can properly import Cost data that includes an Organization reference.


juliusfitzhugh-ccpo (commented):
Looked over the `Costs` table and the `Computed Name` column in `Organization` and they look good (although I haven't worked with the cost data much). I don't see any issues. 

It does seem like there is a lot of extra from the experimentation (uploads). Most of the changes that seem meaningful are towards the bottom search for `/dictionary/x_g_dis_atat_costs.xml` in the files changed tab and most of the changes seem to be from that file downward. From what I remember, I think adding a table adds about ~20-30 file changes.

PR #78, opened 2022-06-09
AT-7452: Define Costs Table Schema

jeffsegal-ccpo:
Note: this is a clean-up version of a previous PR: https://github.com/dod-ccpo/atat-snow/pull/77

Added a new Table name Costs which will contain the cost data we receive from the CSPs. Also added a Computed Name column to Organizations & made it the Display column for the Table so that we can properly import Cost data that includes an Organization reference.

- Adds table `Costs`
- Add column `Computed Name` to table `Organization`
  - pulls from `Organization Name` or `DISA Organization` per business rules

Ticket: AT-7452


jasonburkert-ccpo:
The new table `Costs`

image

The new column `Computed Name`

image

Example of `Computed Name`

image

jeffsegal-ccpo:
> The new table `Costs`
> 
> image
> 
> The new column `Computed Name`
> 
> image
> 
> Example of `Computed Name`
> 
> image

Thanks for documenting all of that - I should have provided those myself.

jasonburkert-ccpo (approved):
Looks good.  Examined in a running environment.

I wonder whether column `Year-Month` should be two columns instead.  I think we should try it this way and reevaluate if necessary.

PR #91, opened 2022-07-05
Add column Schedule Text to table Funding Plan

jasonburkert-ccpo:
This adds a calculated value column to assemble a variable number of funding increments that are part of a funding plan into the desired format.  This will be used on the IFP document.

Adds the following columns:
- `Funding Plan.Schedule Text`

image

Note: This is not the entirety of the ticket implementation, only a necessary component.

Ticket: AT-7513


jasonburkert-ccpo:
image

f600233d1b154d507b782f84604bcb12/update/sys_dictionary_x_g_dis_atat_funding_plan_schedule_text.xml
```diff
@@ -0,0 +1,90 @@
...
+	for (var i = 0; i < increments.length; i++) {
+      var gr = new GlideRecord('x_g_dis_atat_funding_increment');
+      gr.addQuery('sys_id=' + increments[i]);
+      gr.query();
+      while(gr.next()) {
+        // format like 'Funding Increment #X:\nX QTR FYXX - $XX'
+        var line = 'Funding Increment #' + gr.getValue('order') + ":" + newline;
+        line += gr.getValue('description') + ' - ' + gr.getDisplayValue('amount');
+        ouputArray.push(line);
+      }
```
juliusfitzhugh-ccpo, line 21: You already mentioned you would do this, but I already made the comment. I think this is how it would be done using `.get()` from glideRecord.

```suggestion
      gr.get(increments[i]);
        // format like 'Funding Increment #X:\nX QTR FYXX - $XX'
        var line = 'Funding Increment #' + gr.getValue('order') + ":" + newline;
        line += gr.getValue('description') + ' - ' + gr.getDisplayValue('amount');
        ouputArray.push(line);
```

juliusfitzhugh-ccpo (approved):
Looks good, and works as expected. Only left a small suggestion, which you already mentioned you'd do.

PR #92, opened 2022-07-06
AT-7458: Implement Cost Receiver Flow

jeffsegal-ccpo:
This PR adds the capability to pull CostResponse objects from the GET /cost-jobs API (and ultimately from the CostResponse SQS Queue) and populate the SNOW Costs table with the data contained therein. It's composed of the following constructs:

- Action: ATAT Secure GET Request - performs a GET to a HOTH API endpoint
- Action: Persist Costs Payload - persists a JSON payload containing an array of CostResponse objects to the Costs table
- Subflow: ATAT Get Cost Responses - Uses the above Actions to issue a request to the HOTH API and persist the response

image

image

image

Full code of `Persist Costs Payload`:

```js
function addCostRecord(result, csp, organization, portfolio, taskOrder, clin, isActual) {
  var gdt = new GlideDateTime();
  gdt.setDisplayValue(result.month, "yyyy-MM");
  
  // Check to see if we already have a record for this Task Order / CLIN / Year-Month Tuple
  var cost = new GlideRecord('x_g_dis_atat_costs');
  cost.addQuery("task_order_number", taskOrder.task_order_number);
  cost.addQuery("clin", clin.clin_number);
  cost.addQuery("year_month", gdt.getDate());
  cost.query();
  
  var isUpdate = false;
  if (cost.hasNext()) {
    cost.next();
    isUpdate = true;       
  }
  else {
    cost.newRecord();
    isUpdate = false;       
  }

  cost.clin = clin.clin_number;
  cost.csp = csp.sys_id;
  cost.is_actual = isActual;
  cost.organization = organization.sys_id;
  cost.portfolio = portfolio.sys_id;
  cost.service_agency = organization.service_agency;
  cost.task_order_number = taskOrder.task_order_number;
  cost.value = result.value;
  cost.year_month = gdt;
  
  if (isUpdate) {
    cost.update();
  }
  else {
    cost.insert();
  }
  
}

function handleCostObject(costObject) {
  var csp = new GlideRecord('x_g_dis_atat_cloud_service_provider');
  csp.get("name", costObject.content.request.targetCsp.name);

  var portfolio = new GlideRecord('x_g_dis_atat_portfolio');
  portfolio.get("csp_portfolio_id", costObject.content.request.portfolioId);
  
  var organization = portfolio.acquisition_package.organization;
    
  var taskOrders = costObject.content.response.task_orders;
  for (var i = 0; i < taskOrders.length; i++) {
    var taskOrder = taskOrders[i];
[...]

PR #95, opened 2022-07-07
AT-7458: Part Duex

jeffsegal-ccpo:
Attempt to recreate artifacts from https://github.com/dod-ccpo/atat-snow/pull/92 in a clean branch. Please refer there for summary info.


juliusfitzhugh-ccpo (approved):
Looks good and the issues in PR #92 were addressed. I tested the main actions with dummy data and the records were created. 
CreatedCostRecords_AT-7458


Should `num_creates` and `num_updates` be outputs of the `Persist Costs Payload` action? Right now they are only outputs of the step. If that was intended then ignore this. 
NoOutputs_AT-7458

jeffsegal-ccpo:
> Should `num_creates` and `num_updates` be outputs of the `Persist Costs Payload` action? Right now they are only outputs of the step. If that was intended then ignore this.

I mainly included those so we'd have some basic debugging capacity without digging into logs and such. If we need to actually use those in a greater flow (say, the Orchestration Flow that will call this one), then we can make that change there.

PR #98, opened 2022-07-12
Feature/AT-7747: Added Orchestration to drain CostResponse Queue

jeffsegal-ccpo:
**Note**: I mixed up the ticket numbers in my branch - the ticket is actually AT-7747.

* Added a new Flow: `ATAT Cost Response Orchestrator`, which invokes `ATAT Consume Cost Responses` every 30m
* Updated `ATAT Consume Cost Responses` to:
  * Repeat `GET /cost-jobs` query until `[]` is returned or 25 attempts have been made consecutively
  * Output variables on the number of records it created & updated
 

image

image


jasonburkert-ccpo:
Resolved the minor conflicts w/ GH

jeffsegal-ccpo:
> One thing that seemed a bit off was the condition input but appears to work. Should it be like this? It was the same in the sandbox instance as well. 

Nope, good catch. Just pushed a fix: 

image

jeffsegal-ccpo:
> Also, did we want to set the config records for the subflow?

I don't think that buys us much since we have to hard-code it per environment.

juliusfitzhugh-ccpo:
Thanks for answer my questions above. There are two other things that @jasonburkert-ccpo found that he wanted me to mention for him. 

Should the log in the flow have the count data pills? The outputs are there, but the data pills were never added, unless that was intentional. 
LogNoPills_AT-7747

Also, these two look similar to the last fix you made with the data pill not showing the information.
AssignmentValuesMissing_AT-7747

jeffsegal-ccpo:
> Thanks for answer my questions above. There are two other things that @jasonburkert-ccpo found that he wanted me to mention for him.
> 
> Should the log in the flow have the count data pills? 

Thanks, the data pills were definitely there at one point but got garbled along the way. About to push a fix.

juliusfitzhugh-ccpo (approved):
Looks good and ready to go, despite the hard time SNOW gave. Also, thanks for setting an example of the do/while construct.

PR #108, opened 2022-07-22
Added Vue App to ATAT Scoped App

jeffsegal-ccpo:
Re-created Vue App Scripted REST API within ATAT. Accidentally did work as the `admin` user, so the commits all come from that identity.

Scripted REST API for Vue App now added to `ATAT`: 

image

Note that I did not include the menu navigation items since they are likely superfluous. I also have not optimized the scripts yet - that should be a follow-up task.


juliusfitzhugh-ccpo:
@jeffsegal-ccpo will the `index-html` file under `UI Pages` be added in a different PR?

jasonburkert-ccpo (approved):
Viewed the three new REST resources in my PDI.  Confirmed they were transferred from another environment with minimal intended changes.

jasonburkert-ccpo:
The magic number variable `currentRecord` in the scripts for each REST resource contains the `sys_id` of that REST resource.  We'll improve this over time.

When configured with the necessary assets for the Vue application, several files are attached directly to each of the three REST resources.

image

jasonburkert-ccpo:
@juliusfitzhugh-ccpo It appears it has been added under a different name.  I didn't see it before.  Thanks for asking.
image

> @jeffsegal-ccpo will the `index-html` file under `UI Pages` be added in a different PR?

PR #109, opened 2022-07-25
Add Portfolio table security

juliusfitzhugh-ccpo:
Add security controls to the `Portfolio` table to provide access for HaCC admins and Portfolio Managers.

### Create an advanced BR (script)
- `Portfolio - Deny empty managers list` - prevents updating to an empty mission owners `list`
  - toast message pop up (server-side UI)
  - provides system logging for when an update is prevented or reverted to previous value
  - if a record is created without any mission owners, default is for the current user to be provided as a MO
    - unless the user has a role (or is in a group) having permissions of a MO this is probably okay
```javascript
 // Toast message when updating the record in the server-side UI
var message = "Not allowed to update 'Mission Owners' to be an empty list.";
gs.addInfoMessage(message);

// System log
var portfolio = current.name + " portfolio (Sys ID: " + current.sys_id + ") ";
message += " The " + portfolio + " Mission Owners list was not updated ";
message += "with the current value."; 
gs.warn(message);

if (previous.mission_owners.toString() === "") {
  // System log 
  // if the portfolio was created without any MO assigned
  var secondMessage = "No Mission Owners were added to the " + portfolio;
  secondMessage += " during creation. The current user ";
  secondMessage += gs.getUserName() + " (Sys ID: " + gs.getUserID() + ") ";
  secondMessage += "is being added as a default. Update with intended MO.";
  gs.warn(secondMessage);
  
  current.mission_owners = gs.getUserID();

} else {
  // This intentionally updates the mission owners to the previous value.
  // This is done to allow other columns to be updated instead of a full
  // abort which prevents any columns from updating. 
  gs.info("The " + portfolio + " was updated with the previous Mission Owners list.");
  current.mission_owners = previous.mission_owners;	
}
	
```
SystemLogs_DenyEmptyMangersList_AT-7750

### Increase `Portfolio` table security with roles and ACLs
- add advanced access control for allowing `Mission Owners` on a specific portfolio and `HaCC Admins` on a global scope - add `HaCC Admin` role - inherit from `.portfolio_user` role
- create global group with `HaCC Admin` role
AbelUserAccess_AT-7750
HaccAdminAccess_AT-7750
MissionOwnerAccess_AT-7750
Portfolio_AT-7750
 

```javascript
// access control script
var aclObject = { /* access properties */}
anser = hasPermission(aclObject)

// helper
function listContains(item, list) {
    return new global.ArrayUtil().contains(list, item);   
}

// main function to determine permissions
function hasPermission(aclObject) {
    var accessList = aclObject.accessList.toString().split(",");
    var userHasPermission = listContains(aclObject.userSysId, accessList);
    var matchedRoles = aclObject.userRoles.filter(function(userRole) {
		return listContains(userRole, aclObject.allowedRoles);
	});
    var userHasRole = matchedRoles.length > 0;


    if (userHasPermission || userHasRole) {
        return true;
    }

    return false;
}

```

### Update `Portfolio` table
- add reference to `Users` for `Mission Owners` column 
- remove/revert `Portfolio Managers` column
UpdatedPortfolioTable_AT-7750

### For Testing
- Setup a global script to impersonate users
- See ticket for background script used for testing
```javascript
var ImpersonateUser = Class.create();
ImpersonateUser.prototype = {
    initialize: function() {
    },
    impersonate: function(user) {
      var impUser = new GlideImpersonate().impersonate(user);
    },
    type: 'ImpersonateUser'
};
```

Ticket: AT-7750


f600233d1b154d507b782f84604bcb12/README.md
```diff
@@ -16,3 +16,12 @@ If you find yourself unable to import your repository due to the presence of fil
   4. Run `git add -A`
   5. Run `git commit`
   6. Run `git push`
+
```
jasonburkert-ccpo, line 19: didn't we move this file?

jasonburkert-ccpo (approved):
Test script behaves as intended in PDI, executed with three users with different access to Portfolio table.  Nice work!

PR #110, opened 2022-07-26
Add API Gateway ID

kylelaker-ccpo:
This adds support for passing the `x-apigw-api-id` header via a new configuration in the `ATAT Configuration` table. Additionally, the `GET` and `POST` actions have been unified to accept an `ATAT Configuration` reference and to then handle the values; there was a behavior difference there before in that one took the reference and another took a string.


zachclark-ccpo:
@kylelaker-ccpo could you screenshot the flows/scripts you changed?

kylelaker-ccpo:
> @kylelaker-ccpo could you screenshot the flows/scripts you changed?

image

image

image

image

image

image

image

image

image

image

juliusfitzhugh-ccpo (changes requested):
Nice work, thanks for cleaning things up and making things consisitent. Overall I think this will work, but did find a few things.

### Working nicely
The `ATAT Cost response Orchestrator` flow is working which uses the `ATAT Consume Cost Response` subflow and the `ATAT Secure GET Request` action, so the subdomain does work in the sandbox instance.
CostResponseInputs
CostResponseFlowSuccessCondensed
CostResposneLogs

### Minor updates
The following subflows need to be *published* to ensure the new `api-id` input shows up on the parent flows that use them (e.g., ATAT Cost Orchestrator flows)
- `ATAT Consume Cost Responses` subflow
- `ATAT Send Cost Request` subflow

### Tried trouble shooting, but couldn't get to the source
The biggest issue i've found is the `ATAT Secure POST Request` action failing and causing other flow/subflows to fail as well. Below is the issues I'm seeing and a small fix for one
- first fix is the baseURL in the REST step need to provide the `value` of the record for the step to use UpdatedBaseURL
- some unknown error that I've not seen before but seems like there is some kind of parsing error
- Also, not sure if this is related or a side effect of something else but the region shows up in the `Acquisition package`, but wanted to point it out at least
SecurePostRequestError

I discarded all the changes I had on your branch to clean things back to how you had them.

jeffsegal-ccpo:
Really appreciate the thorough and organized review @juliusfitzhugh-ccpo !

I have seen the variables getting swapped like that before. If I recall, it was usually caused by sending null values for the variable in question but obviously there's some silly array reference stuff happening. I'll reach out to Nishant when he's available.

kylelaker-ccpo:
> ### Minor updates
> The following subflows need to be _published_ to ensure the new `api-id` input shows up on the parent flows that use them (e.g., ATAT Cost Orchestrator flows)
> 
> * `ATAT Consume Cost Responses` subflow
> * `ATAT Send Cost Request` subflow

Oops. Okay. So my PDI showed "Activate" and I feel like I've been avoiding "Activate"ing things so I didn't press it. But I think that's actually "Publish" and my PDI is weird.

> ### Tried trouble shooting, but couldn't get to the source
> The biggest issue i've found is the `ATAT Secure POST Request` action failing and causing other flow/subflows to fail as well. Below is the issues I'm seeing and a small fix for one
> 
> * first fix is the baseURL in the REST step need to provide the `value` of the record for the step to use UpdatedBaseURL

Fixed

> * some unknown error that I've not seen before but seems like there is some kind of parsing error

Looks like that's because I missed the same thing for "Resource Path"

> * Also, not sure if this is related or a side effect of something else but the region shows up in the `Acquisition package`, but wanted to point it out at least

As Jeff mentioned, this seems to just sorta happen when something is null/undefined? I don't think it's the cause but if we see it seeming to cause issues we can dive in more.

kylelaker-ccpo:
Thanks for that additional fix, @juliusfitzhugh-ccpo! Not sure why it didn't publish but I appreciate you taking care of it!

juliusfitzhugh-ccpo (approved):
Did another test run of the following flows/subflows/actions and all are working as expected. Had to publish the `ATAT Secure Post Request` but that was all. Thanks for applying the changes to incorporate the custom domains!

Note: The error shown in the `ATAT Consume Cost Responses` subflow is expected since there is a 400 response from the mock server. The parent `ATAT Cost Response Orchestrator` flow also returned an error because of that, but that's also expected. 

### 

- `ATAT Cost Orchestrator` - flow
    - `ATAT Send Cost Request` - subflow
        - `ATAT Secure Post Request` - action (needed to be published)
CostOrchestratorPassed

- `ATAT Cost Response Orchestrator` - flow
    - `ATAT Consume Cost Responses` - subflow
        - `ATAT Secure GET Request` - action
ConsumeCostResponsePass
Response

- `ATAT Generate Description of Work` subflow
    - `ATAT Secure GET Request` - action
GenerateDoW_Pass
AttachedDow

kylelaker-ccpo:
Thanks for the thorough review! I am going to go ahead and merge. This is really exciting that we're able to get this in!

PR #113, opened 2022-07-29
Add support for Operator management

jasonburkert-ccpo:
Add support for Operator management to assist with provisioning operations.

### `Operator` table
- Create `Operator` table
- Create advanced BR to prevent returning `provisioned` column to `false`

### `Portfolio` table 
- Add column `Operators` as `List`

### `ATAT Provisioning Request Flow` flow
- Rename existing flow to `ATAT Provisioning Request Flow` - action name only, not the internal name
- Update flow to account for creating operator records if a successful request
  - This flow will be refactored, so possibly not the final implementation related to operators
  - Operators will not be related to any of the package contacts - so will be removed in the refactor.

Ticket: AT-7749


jasonburkert-ccpo:
Table `Operator`

image

New column in table `Portfolio`

image

zachclark-ccpo (approved):

PR #114, opened 2022-08-02
Update `Provisioning Job` table

juliusfitzhugh-ccpo:
`Provisioning Job` table
- Add `Payload` column - string
- Add choice `NOT_STARTED` for `Status` column
**Note:** `Data Structure` did not seem useful  so went with string. See ticket comment for details.
UpdateProvisioningJobTable_AT-7803

Ticket: AT-7803


jeffsegal-ccpo (approved):
Validated in my PDI, looks good.

PR #115, opened 2022-08-04
Create Acq Package submission flow

juliusfitzhugh-ccpo:
Create a flow that initialize the start and creation of provisioning objects. 

### Create `ATAT Acquisition Package Submission` flow
- trigger when package updates to under review
- create shell portfolio
  - add assigned fields to the shell portfolio
    - `Acq Pkg` - ref
    - `MO` - List
    - `Name` - String
    - `Operators` -  List (same list as package)
    - `Provisioned` - True/False
- add basic error handler
AcqPackageSubmission_AT-7805
ShellPortfolio_AT-7805

### Update `Acquisition Package` table
- add `Operators` column - List

### Update `Portfolio` table
- add `Provisioned` column - True/False

Ticket: AT-7805


jasonburkert-ccpo (approved):
Tested in a PDI.  Behaves as expected.  Nice work!

### Recommendation
It’s being set in the flow, but I’d recommend setting a default value of `false` for column `Portfolio.Provisioned`

image

jasonburkert-ccpo:
Very good.

image

PR #120, opened 2022-08-15
Change column type for Alert.Last Notification Date

jasonburkert-ccpo:
Changes column from `Date` to `DateTime`

Sample Alert in list layout
image

Ticket: AT-7848


zachclark-ccpo (approved):
Looks good.

PR #121, opened 2022-08-15
Persist entries of table Classification Level

jasonburkert-ccpo:
This data is currently persisted in folder `/data-tables/x_g_dis_atat_classification_level.xml`
This PR replaces that file by leveraging a different ServiceNow feature to persist that data.
This PR creates a file `update/sys_metadata_link*.xml` file for each data row.  (for every sys_id in the imported XML, there is a resulting sys_metadata_link*.xml file)

### How to do this (should be instructive for other tables)
- Import file such as `/data-tables`/[x_g_dis_atat_classification_level.xml](https://github.com/dod-ccpo/atat-snow/blob/4e86306925c678371fcd31136f54e93366252bfc/data-tables/x_g_dis_atat_classification_level.xml) into empty table `Classification Level` with feature (right-click column header) _Import XML_

- Select checkboxes for all data rows in the table and use feature (drop-down at bottom of table rows) _Create Application File_

- Commit the newly generated `sys_metadata_link*.xml` files


jasonburkert-ccpo:
Once created these application files can be found via `sys_metadata.list` and filtering accordingly

image

jasonburkert-ccpo:
Once merged, file `/data-tables/x_g_dis_atat_classification_level.xml` can be removed when we're satisfied it is no longer needed.

juliusfitzhugh-ccpo (approved):
Nice work testing this out on the `classification_instance` table. In PDI, the records persist as expected. I've also tried switching to different branches without the changes in this PR and the records persist  until they are deleted. Switching back to the branch with the records (source control application file), and the records are available again.

Appears to be working as expected for persisting table records.

PR #126, opened 2022-08-24
Implemented Evaluate Time Remaining for CLIN Subflow

jeffsegal-ccpo:
Given a CLIN, this will determine if an Alert should be created or updated, depending on how close the current date is to the PoP End Date.

image

To test, create 4 CLINs with the following characteristics:

1. UPCOMING - PoP Start Date is in the future
2. ACTIVE - PoP Start Date is in the past and PoP End Date is at least 60 days from now
3. AT-RISK - PoP Start Date is in the past and PoP End Date is 60 days or less from now
4. EXPIRED - PoP End Date is in the past

Note that `CLIN Status` will be automatically evaluated based on the dates. They should look like this:

image

Pass each into the Subflow and verify the following expectations:
1. UPCOMING - nothing happens
2. ACTIVE - nothing happens
3. AT-RISK - a new Alert is created (fields should be validated appropriately)
4. EXPIRED - nothing happens

Next, manually extend the AT-RISK CLIN's PoP End Date by a few days, but less than 60 from now. Pass it into the Subflow again and verify that the `Threshold Violation Amount` reflects an updated number of days remaining and `Active` is still set to `true`.

Finally, manually extend the AT-RISK CLIN's PoP End Date to more than 60 days from now. Pass it into the Subflow again and verify that the `Threshold Violation Amount` reflects an updated number of days remaining and `Active` is now set to `false`. This simulates a Mission Owner funding a follow-on CLIN.


juliusfitzhugh-ccpo (approved):
Nice work with the low code approach! The subflow works as expected, thanks for the testing instructions. Think you need to update the `.secrets.baseline` both otherwise seems good to go.

There was one part I did not fully get, you mentioned a bug in the annotation, could you explain that a bit more?
Bug

PR #129, opened 2022-08-26
AT-7854: Final tweaks

jeffsegal-ccpo:
Updated Get Effective Follow-on CLIN to return the input CLIN if no follow-on is found.
image

Updated Evaluate Time Remaining for CLIN to calculate Days Remaining based on follow-on CLIN.
image


juliusfitzhugh-ccpo (approved):
Looks good!

PR #134, opened 2022-09-02
Add tables Package Document and Package Document Type

jasonburkert-ccpo:
- Adds table `Package Document` - contains document instances/revisions
- Adds table `Package Document Type` - defines document types required for an acquisition package for a given contract
  - Adds 12 reference rows for known document types

In the future the generated documents we've been attaching directly to table `Acquisition Package` will be relocated to this new table `Package Document`.

Ticket: AT-7818


jasonburkert-ccpo:
### `Package Document Type` table
A simple Choice field to document types seemed inadequate.  This gives us more flexibility.  These records are controlled as reference data.

This `File Name` column is a preference, with these values matching EDMS with the exception of IGCE.  Interestingly the checklist is the only one to use a hyphen rather than a space.

`EDMS Document Type` values came from our early integration research.

image

### `Package Document` table
We may add a sequence number or similar in the future for revisions.  Get max sequence # for combination of `Acquisition Package` + `Document Type` and increment for composite key to insert new row for each round of review.

The `File` column is a binary attachment.

image

juliusfitzhugh-ccpo (commented):
Nice job, I can see how a `Choice` type would not serve justice to things. There is more background and context than I realized with these documents, but these table help show that beside just saying `document XX`. 

Few questions (see ticket comment for expanded question and screenshoots):
- Do we need a reference to the documents from the package?
- Is the `File` column on `Package Document` going to behavior similar when we attached files to the record?

jasonburkert-ccpo:
> * Do we need a reference to the documents from the package?

No, because each `Package Document` row has a reference to the `Acquisition Package` to which it belongs.  This is similar to how `Portfolio` is related to `Acquisition Package`.  If we turn on the display of the related list for Package Documents in the Acquisition Package form, we can see each document... (in fact, this is probably a change I should make in this PR)

image



> * Is the `File` column on `Package Document` going to behavior similar when we attached files to the record?

This is a very good question.  I was assuming that a column of type `File Attachment` would work the same as any attachment on a record, but it does seem to be slightly different.  I'll experiment.

juliusfitzhugh-ccpo (approved):

jasonburkert-ccpo:
I haven't had much time to explore the _File Attachment_ column and how it works programmatically and how that compares to what we're doing today on `Acquisition Package` records.  We can try using the new `File` column, if it doesn't work we can attach files directly to the record which should be essentially identical to what we're doing today.

PR #142, opened 2022-09-14
Remove all under 'Navigation' in ServiceNow Studio

jasonburkert-ccpo:
Removes all under `Navigation` in ServiceNow Studio
- Navigation > Application Menus
- Navigation > Modules
- Navigation > Application Menus (Mobile)
- Navigation > Modules (Mobile)

No need to control and maintain these files as we don't use the navigation components within ServiceNow.

### Note
While it might be expected that this change set is largely deletions, there is also a significant amount of additions.

I removed about 165 application files.  These can be see in _Files changed_ under `f600233d1b154d507b782f84604bcb12/update`

The system created all the entries under `f600233d1b154d507b782f84604bcb12/author_elective_update` for processing deletes on updates.

### Explanation
Performed some testing.  When creating new tables we should uncheck _Create module_ which is checked by default.  When checked this causes the creation of several navigation related components.

This is what we've been doing when creating tables.

image

Instead, this is what we should do going forward.

image


zachclark-ccpo:
My only concerns are:

- Do we need to keep doing this mass cleaning or is there an option we can enable to prevent these Navigation artifacts from being created?
- Should the developer be responsible with removing any Navigations when creating a PR after this is merged?

jasonburkert-ccpo:
> My only concerns are:
> 
> * Do we need to keep doing this mass cleaning or is there an option we can enable to prevent these Navigation artifacts from being created?
> * Should the developer be responsible with removing any Navigations when creating a PR after this is merged?

Unchecking `Create module` when creating a table in the future will prevent the creation of these navigation components.

This mass cleaning is only necessary because we created a bunch of navigation components when creating each table because we simply didn't know any better.  They might be useful if we were building interfaces in ServiceNow, but given our use case, we primarily want to define tables and forms for data entry.

jasonburkert-ccpo:
We don't necessary need to commit this PR, but I at least wanted to introduce the idea that we don't need the Navigation components, and should try to prevent their creation when adding tables.

juliusfitzhugh-ccpo:
I'm fine with removing the 'Navigation' in the app. I think there are ways for each user to add favorites to help with navigation if needed. 

I think this is okay currently, but we should probably avoid doing mass cleanings the more solidified the scoped app becomes. While I think it's good to clean up unnecessary things, the behavior of SNOW does not give me reassurance that something behind the scenes will not cause issues.

juliusfitzhugh-ccpo (approved):
I thought we might use navigation more, but I guess not. @jasonburkert-ccpo the `.secrets.baseline` and `checksum` need to be updated if you still want to merge this PR.

jasonburkert-ccpo:
### Before this PR

image

### After this PR (no Navigation components)

image

PR #145, opened 2022-09-19
changing Emergency Declaration Column to match other YES/NO column co…

robertmccardell-ccpo:
Updating Emergency Declaration Column in Project Overview to match the YES/NO conventions of other choice columns in DISA


juliusfitzhugh-ccpo (approved):
Looks good.

PR #149, opened 2022-09-23
Add column Portfolio Status to Portfolio Table

zachclark-ccpo:
# Overview
Added a new column `portfolio_status` to the Portfolio Table.
It is a Choice with the following options:
image

Ticket: [AT-7973](https://ccpo.atlassian.net/browse/AT-7973)


juliusfitzhugh-ccpo (approved):
Looks good!

PR #150, opened 2022-09-23
Update Task Order and CLIN Status options

zachclark-ccpo:
# Overview
Updated the options for Choice type columns to match the documentation [here.](https://ccpo.atlassian.net/wiki/spaces/AT/pages/2460385308/Portfolio+Task+Order+and+CLIN+Statuses+Business+Rules)

## CLIN Status
`x_g_dis_atat_clin.clin_status`
image

## Task Order Status
`x_g_dis_atat_task_order.task_order_status`
image


f600233d1b154d507b782f84604bcb12/update/sys_dictionary_x_g_dis_atat_clin_clin_status.xml
```diff
@@ -7,19 +7,9 @@
...
-	if (current.pop_end_date < now) {
-		return 'EXPIRED';
-	}
-	else if (current.pop_start_date > now) {
-		return 'UPCOMING';
-	}
-	else if (current.pop_end_date < twoMonthsFromNow) {
-		return 'AT_RISK';
-	}
-	return 'ACTIVE';
```
juliusfitzhugh-ccpo, line 22: Do we want this removed or will this be done somewhere else? I thought the idea was to have this be automatically updated based on the PoP dates, but that might be scope of another ticket.

juliusfitzhugh-ccpo (approved):
Looks good, one small comment, but might be out of scope for this ticket.

f600233d1b154d507b782f84604bcb12/update/sys_dictionary_x_g_dis_atat_clin_clin_status.xml
zachclark-ccpo: Since the options now are:

- On-Track
- Expiring PoP
- Funding At_risk
- At-Risk
- Option Pending
- Option Exercised
- Expired
- Delinquent
I think a full ticket could be made to support all the options, if there isn't one already made.

juliusfitzhugh-ccpo: Fair it does seem more involved then the ticket and should be captured.

PR #154, opened 2022-09-30
Add `Populate Agencies from CSS` flow

juliusfitzhugh-ccpo:
### Add `Agency` table
AddAgencyTable_AT-7965

### Update `Organization` table 
- Add `Agency` column
- Remove `Service/Agency` column
UpdateOrganizationTable_AT-7965

### Update business rules
- Org - Only DISA has a 'DISA org.'
- Org - Org. Name null if Agency is DISA

### Add `Populate Agencies from CSS` flow
- create `Make A Request` action (attempt to make generic for other requests)
  - Uses `ATAT Configuration` records for inputs (when inputs are typed action fails) 
- create `Insert CSS Agencies` action
  - `Format Agencies` script
  - `Insert Agencies` script  

PopulateAgenciesCSS_AT-7965
MakeARquest_AT-7965
InsertCSS_AT-7965

### Update `ErrorHandler` script include
- Add `DuplicateRecord` error


jeffsegal-ccpo (changes requested):
Good stuff. A couple of items: 

1. Make sure we can configure the `mutualAuth` setting in `Make a Request`
2. In `Insert Agencies`, we shouldn't treat the existence of a matching agency as an error since it'll be the most common occurrence by far. Better to silently ignore those records. Also I'd either remove the `gs.debug` line after insertion or format it so that we know what it is if we were looking in the global log view.
3. The JWT can still be an input to the `Make a Request` Action but we should add a custom script for it to use `getCssJwt`.

juliusfitzhugh-ccpo:
@jeffsegal-ccpo for number 3 above, are you suggesting we add an additional action before the `Make A Request` action that uses the `getCssJwt`? i just want to ensure I'm interpreting it correctly.

jeffsegal-ccpo:
> @jeffsegal-ccpo for number 3 above, are you suggesting we add an additional action before the `Make A Request` action that uses the `getCssJwt`? i just want to ensure I'm interpreting it correctly.
> 
> Or just the script for that particular input?

This is what I mean:

image

juliusfitzhugh-ccpo:
Changes from feedback

- Remove error and debug
- Add dynamic mutual auth input
MutualAuthInput_AT-7965
- Use `getCssJwt` script for input
JWT-input_AT-7965

jeffsegal-ccpo (approved):
Good to go!

PR #156, opened 2022-10-03
Feature/AT-7963: Populate DITCO CS CSS IDs

jeffsegal-ccpo:
Added a new Flow, `Populate DITCO Contract Specialist CSS IDs`, which iterates through the new DITCO CS Table and attempts to update the `CSS ID` column for each entry. It will invoke different APIs depending on whether the Person is a KO or not.

image
image

### Modified `Request` to accept a Content-Type
image

### Modified `getCssJwt` to attempt to look up a CSS Endpoint if not passed in:
image

### Other Changes
* Added `DITCO Contract Specialist` Table
* Added `DITCO KO` column to `Acquisition Package`
* Renamed `Primary Reviewer` column to `DITCO CS` and changed reference from `User` to `DITCO Contract Specialist`
* Refactored `Primary Reviewer` reference in `Pkg - Primary Reviewer Status update` business rule
image


f600233d1b154d507b782f84604bcb12/dictionary/x_g_dis_atat_acquisition_package.xml
```diff
@@ -27,35 +28,35 @@
             </choice>
         </element>
         <element label="Period Of Performance" max_length="32" name="period_of_performance" reference="x_g_dis_atat_period_of_performance" type="reference"/>
-        <element label="Primary Reviewer" max_length="32" name="primary_reviewer" reference="sys_user" type="reference"/>
+        <element label="DITCO CS" max_length="32" name="primary_reviewer" reference="x_g_dis_atat_ditco_contract_specialist" type="reference"/>
```
juliusfitzhugh-ccpo, line 31: I noticed in my PDI the `DITCO CS` column does not show up as expected but appears there. However, the column shows up in DSP. The under lying column is still named `primary_reviewer` is that intended?

f600233d1b154d507b782f84604bcb12/update/sys_script_include_9738ccda974611106fa8b4b3f153af13.xml
```diff
@@ -17,6 +17,24 @@
...
+	try {
+		if (!endpoint) {
+			cssEndpoint.addQuery("name", "cssBaseUrl");
+			cssEndpoint.query();
+			if (!cssEndpoint.next()) {
+				gs.error("No endpoint found in ATAT Configuration.");
+				return null;
+			}
+			else {
+				endpoint = cssEndpoint.value + "/api/Auth/cert";
```
juliusfitzhugh-ccpo, line 32: Nice, thanks for making it easier to get a token!

f600233d1b154d507b782f84604bcb12/update/sys_script_eacf8da0970d91102b1ef0b0f053af4e.xml
```diff
@@ -36,7 +36,7 @@
 	var aclObject = {
 		userSysId: gs.getUserID(),
 		userRoles: gs.getUser().getRoles().toString().split(","), 
-		accessList: current.primary_reviewer.toString().split(","),
+		accessList: current.primary_reviewer && current.primary_reviewer.user ? current.primary_reviewer.user.toString().split(",") : null,
```
juliusfitzhugh-ccpo, line 39: If it was intended to leave the underlying column name as `primary_reviewer` this is fine, but if not it would need to be updated.

f600233d1b154d507b782f84604bcb12/update/sys_script_include_5d76f38e970611106fa8b4b3f153afc4.xml
```diff
@@ -39,6 +39,9 @@ Request.prototype = {
 		if (req.token) {
 			request.setRequestHeader("Authorization", "Bearer " + req.token);
 		}
+		if (req.contentType) {
+			request.setRequestHeader("Content-Type", req.contentType);
+		}
```
juliusfitzhugh-ccpo, line 44: Good catch we need that for actual POST for sure.

juliusfitzhugh-ccpo (commented):
Flow looks good, thanks for the improvements. Few comment/question, but otherwise it should be ready to go. 

Just a thought, but would we want to make the two requests in parallel for the flow?

f600233d1b154d507b782f84604bcb12/dictionary/x_g_dis_atat_acquisition_package.xml
jeffsegal-ccpo: Yeah, unfortunately you can't change column names, just labels. 🙄

f600233d1b154d507b782f84604bcb12/update/sys_script_eacf8da0970d91102b1ef0b0f053af4e.xml
jeffsegal-ccpo: There were two changes - the name AND the type, from a `Reference` to `User` to a `Reference` to `DITCO Contract Specialist`.

f600233d1b154d507b782f84604bcb12/dictionary/x_g_dis_atat_acquisition_package.xml
jeffsegal-ccpo: Fair - I was considering leaving the label as-is too, but figured I'd compromise and change the label without actually destroying the whole field.

juliusfitzhugh-ccpo (approved):
@jeffsegal-ccpo leaving the decision up to you.

PR #157, opened 2022-10-04
Update `Costs` table to reference Agency table

juliusfitzhugh-ccpo:
Update `Costs` table columns
- Add `Agency` column
- Remove `Service/Agency` column

Ticket: AT-7965


jasonburkert-ccpo (approved):
Verified changes in PDI

PR #168, opened 2022-10-14
Disable two Acquisition Package business rules

jasonburkert-ccpo:
Disables the following business rules
- `Pkg - Mission Owner status update`
- `Pkg - Primary Reviewer status update`

These rules need to be updated and re-evaluated due to recent events.


juliusfitzhugh-ccpo (approved):
Thanks for disabling these.

PR #173, opened 2022-10-24
Begin replacing 'ATAT Configuration' with System Properties

jeffsegal-ccpo:
### Prototyping replacing `ATAT Configuration` Table with System Properties

I began by choosing `ATAT Secure POST Request` as the starting point, also renaming it to `HOTH API POST Request`. I removed extraneous inputs, such as Region, which would instead be obtained from System Properties which I separately created, e.g.:

New code:

```js
httpRequestData.setRegion(gs.getProperty("x_g_dis_atat.aws.region"));
```

New property:

image

The same was done for all other system-wide inputs, including:
* `Region` was replaced with property `x_g_dis_atat.aws.region`
* `API Gateway ID` was replaced with property `x_g_dis_atat.aws.hoth.apiGatewayId`
* `HOTH Base URL` was replaced with property `x_g_dis_atat.aws.hoth.baseUrl`
* `AWS Credential` was replaced with property `x_g_dis_atat.aws.credential.sys_id`

### Downstream Flows

There are several flows which used the former `ATAT Secure POST Request` which also had to be changed:

* `ATAT Generate Description of Work`
* `Generate IGCE`
* `ATAT Send Cost Request`
* `ATAT Provisioning Request`

Each of these were similar, removing extraneous inputs and hard-coding resource paths as appropriate, e.g.:
image

 Ticket: AT-8172


f600233d1b154d507b782f84604bcb12/update/sys_hub_action_type_definition_921e87a8978011101e0835dfe153afb4.xml
```diff
@@ -1583,10 +1583,12 @@ Request Body:
   
   
   var headers = signedRequest.getHeaderMap();
-  outputs.host_header = base_url;
+  outputs.base_url = base_url;
+  outputs.host_header = "api.atat.dev";
```
jasonburkert-ccpo, line 1587: Should this also become a system property?

jasonburkert-ccpo (approved):
This looks good.  Left a couple of questions in comments above.

f600233d1b154d507b782f84604bcb12/update/sys_hub_action_type_definition_921e87a8978011101e0835dfe153afb4.xml
jeffsegal-ccpo: Ah thanks for catching that. I was troubleshooting and hard-coded that. Will revert that back to being parameterized.

jeffsegal-ccpo:
> Question. Do we care about the disparity between the renamed label name, and the internal name? The internal name reflects the old name, which could complicate future troubleshooting. I'd rather this wasn't the case, but think it's fine as long as we do it with eyes wide open.
> 
> image

Fair question but as a general rule, I think I'd prefer to change labels to something more meaningful rather than be stuck with the original names forever.

PR #177, opened 2022-10-31
Use ServiceNow system properties instead of 'ATAT Configuration' table

jeffsegal-ccpo:
Refactored all remaining `ATAT Configuration` usages _except_ `cssJwt`, which is managed automatically. Includes updates to the following:

- Scripted REST APIs
  - EDA
  - G-Invoicing
- Flows
  - Populate DITCO Contract Specialist CSS IDs
  - ATAT Acquisition Package Submission
  - Populate Agencies from CSS
- Subflows
  - Create CSS Contract Record
  - EDMS: Upload Documents
- Actions
  - Construct CSS Contract Record
  - EDMS: Create Post Award Folder
  - Make A Request

Ticket: AT-8187

PR #178, opened 2022-10-31
Add prefixes DAPPS and ATAT to table labels

jasonburkert-ccpo:
These prefixes help differentiate table use and purpose.

### DAPPS prefix
These tables are used pre-task order award.  These support package preparation in the run up to task order award.

### ATAT prefix
These tables are used post-task order award.  These support provisioning of cloud environments, and management and reporting on those environments.

### DELETE prefix
We anticipate removing these tables in the near term due to #177

### Tables worthy of further discussion
It's clear on which side most tables belong, but some are worthy of additional debate for clarity.  I've given them prefixes in this PR, but speak up if you think they should be different.

- `Cloud Service Provider` - this is referenced in both parts of the process, however, much more critical use on the ATAT side
- `Task Order` - currently mixed use and there is an outstanding task to separate the purposes by creating a second table
- `Agency` - is populated by an external system since #154

image


jeffsegal-ccpo (changes requested):
This will be useful. Only issue is that `ATAT Configuration` can't be deleted until we also refactor the `cssJwt` key to some other construct.

jasonburkert-ccpo:
> This will be useful. Only issue is that `ATAT Configuration` can't be deleted until we also refactor the `cssJwt` key to some other construct.

To be clear I've not removed the table in this PR, only given it the _DELETE_ prefix to communicate intent

jeffsegal-ccpo (approved):

PR #184, opened 2022-11-04
Add table 'DAPPS:Current Environment System Documentation'

jasonburkert-ccpo:
This table acts as a bucket to categorize n number of attachments

Ticket: AT-8244

PR #185, opened 2022-11-04
Add 'DAPPS:Current Environment Migration Documentation'

jasonburkert-ccpo:
This table acts as a bucket to categorize n number of attachments

This has the same form and function of #184 but merely represents a different bucket/category

Ticket: AT-8245


jasonburkert-ccpo:
### New table

image

### Form layout displays attachments

image

PR #187, opened 2022-11-07
Normalized Flows, Subflows and Action names

jeffsegal-ccpo:
Added the following prefixes to help us sort our various Flow Designer constructs:

- COST
- SF
- DOCGEN
- PROV
- CSS
- EDMS
- ALERTS
- CSS
- UTIL

Nearly all constructs were updated, except a handful which the UI would not allow for some unknown reason.


juliusfitzhugh-ccpo (approved):
Naming looks good and helps sort the different flows/subflows/actions.

PR #188, opened 2022-11-08
Create application files for Military Rank records

juliusfitzhugh-ccpo:
Create application files for `Military Rank` records to be tracked by source control.

Ticket: AT-8140


jasonburkert-ccpo (approved):
Looks good.  I see one `sys_metadata_link` file for each of the 145 records in the XML export that we intent to replace.

PR #189, opened 2022-11-08
Add table 'DAPPS.Current Environment Instance'

jasonburkert-ccpo:
- Adds table `DAPPS.Current Environment Instance`
- Adds table `DAPPS.Region`
  - includes reference records
  - records grouped by CONUS/OCONUS

Ticket: AT-8246


jasonburkert-ccpo:
### New Region table
These are generic regions with a USG/DoD focus, and not belonging to a specific CSP.

image

jasonburkert-ccpo:
This was helpful to work on details.  But further discussion with team members makes it clear I should take what we've learned and start in a fresh branch.

PR #191, opened 2022-11-09
Create table schema to store Current Environment data

jasonburkert-ccpo:
- Adds table `Region` including reference data

Ticket: AT-8246


jasonburkert-ccpo:
Replacing with https://github.com/dod-ccpo/atat-snow/pull/195 and others.

PR #192, opened 2022-11-11
Bugfix: AT-8187

jeffsegal-ccpo:
* Fixed a few regressions caused by extra whitespace characters in CSS flows 
* Updated Portal Page name so it looks better in browser tabs 
* Updated `Igce` to take an Acquisition Package record as input since the `getPayload()` function was expecting it rather than an ID already


juliusfitzhugh-ccpo (commented):
Bugs fixed. The only thing is the `CSS: Construct Contract Record` action needs to be published for the updates to happen in the submission flow. 

Also, `secrets.baseline` needs updating.

juliusfitzhugh-ccpo (approved):

PR #199, opened 2022-11-21
Add two system properties to the ATAT application scope

jasonburkert-ccpo:
Adds two system properties
- `x_g_dis_atat.dappsTemplates.justificationAndApproval`
- `x_g_dis_atat.dappsTemplates.marketResearchReport`

These will have respective document templates attached.  These are system level templates to be made available for user download.

Note: The properties are part of the application scope and source controlled (this PR).  However, the actual file attachments and their content are considered instance data by ServiceNow.

This is in support of ticket AT-8338.


jasonburkert-ccpo:
### dappsTemplates.justificationAndApproval

image

### dappsTemplates.marketResearchReport

image

juliusfitzhugh-ccpo (approved):

juliusfitzhugh-ccpo:
Nice idea with using system properties!

PR #200, opened 2022-11-22
Control document templates

juliusfitzhugh-ccpo:
Document Templates
- JWCC MRR
- JWCC J&A


jasonburkert-ccpo (changes requested):
After updating PDI, the two system properties each have attachments with the expected name, but upon downloading the attachments and inspecting the files they appear to be empty.

Looking at the _Files changed_ tab in GH - where is the content of the two files?  It looks like there is metadata for the attachments, but no content.

### Example
These look OK
image
image

But the files are 0 bytes each
```
2022-11-22  12:36 PM                 0 JWCC J&A Template_Template.docx
2022-11-22  12:36 PM                 0 JWCC Market Research Report (Sole Source)_Template.docx
               2 File(s)              0 bytes
```

jasonburkert-ccpo:
@juliusfitzhugh-ccpo going to leave this open for now, but expect we'll close it if no breakthroughs

juliusfitzhugh-ccpo:
@jasonburkert-ccpo I switched to draft since things need to be looked at a little more.

jasonburkert-ccpo:
The J&A and MRR templates have been made available by other means.

PR #201, opened 2022-11-23
Update EDA system property (protocol)

juliusfitzhugh-ccpo:
Update `eda.protocolProfile` system property from value _mauth_ to _eda_


juliusfitzhugh-ccpo:
Might not be needed, so converted to draft for the time being.

PR #204, opened 2022-11-30
Create Generate IFP subflow

juliusfitzhugh-ccpo:
### Create `DOCGEN: Generate IFP` subflow
GenerateIFPsubflow_AT-8377

### Update `Funding Plan` table
- Add an additional `newline` to the `Schedule Text` column for space between increments

Ticket: AT-8377


zachclark-ccpo:
@juliusfitzhugh-ccpo  Might be out of scope for the ticket but.... in the third step: Attach Binary File to Acquisition Package... I know that there is a new table to store Acquisition Package documents, is this step using that table?

juliusfitzhugh-ccpo:
> I know that there is a new table to store Acquisition Package documents, is this step using that table?

@zachclark-ccpo the new table is not being used for storing the documents in the Attach Binary File action. I think the new table is the `Package Document` table, but not sure if thats for all documents or finalized ones. Also, I'm not sure if we figured out how to actually attach them in the record for the `attachment` type column. Ideally it should be simple.

zachclark-ccpo:
> > I know that there is a new table to store Acquisition Package documents, is this step using that table?
> 
> @zachclark-ccpo the new table is not being used for storing the documents in the Attach Binary File action. I think the new table is the `Package Document` table, but not sure if thats for all documents or finalized ones. Also, I'm not sure if we figured out how to actually attach them in the record for the `attachment` type column. Ideally it should be simple.

Hmmm Im sure we can just update the UTIL if we need too, preferably in a separate ticket.

zachclark-ccpo (approved):
Flow looks good, make sure to update the checksum before merging

PR #207, opened 2022-12-01
Update reference data in 'Service Offering' table

jasonburkert-ccpo:
- Removes some options from column `Service Offering Group`
- Adds other options to same
- Adds eight rows of reference data to `Service Offering` table

Ticket: AT-8415


jasonburkert-ccpo:
### Updated options for column Service Offering Group

image

jasonburkert-ccpo:
### New Reference data rows

image

PR #215, opened 2022-12-02
Feature/AT-8375: Extract Eval Plan data

jeffsegal-ccpo:
* Created Eval Plan Script Include to ETL data from SNOW to HOTH schema
* Added DOCGEN: Generate Eval Plan subflow

Tested it with a fully populated Eval Plan object as well as a completely empty one.


juliusfitzhugh-ccpo (approved):
Looks good. Nice use of the script for setting the payload in the subflow. I always forget about it.

PR #214, opened 2022-12-02
Add table 'Selected Classification Level'

jasonburkert-ccpo:
An acquisition package applies to some Classification Levels which the mission owner selects.  These selections help define the acquisition package.

- Adds table `Selected Classification Level`
- Remove column `Acquisition Package.classification_level`
- Updates Relational Model Diagram to reflect the change

Ticket: AT-8389

PR #217, opened 2022-12-05
Add columns to table 'Selected Classification Level'

jasonburkert-ccpo:
Builds on #214 

- Adds nine columns to `Selected Classification Level`
- Updates Relational Model Diagram to reflect the change

Ticket: AT-8389


jasonburkert-ccpo:
### New columns

image

juliusfitzhugh-ccpo (commented):
Looks good, the only thing that might needs updating is the default for the `Users Per Regions` column.  Might need more than 40 for that since it is a JSON string.

juliusfitzhugh-ccpo (approved):

jasonburkert-ccpo:
> Looks good, the only thing that might needs updating is the default for the `Users Per Regions` column. Might need more than 40 for that since it is a JSON string.

Fixed in https://github.com/dod-ccpo/atat-snow/pull/217/commits/ed3cbe2030befc1ab66fcdfc389999aecd9fe022

PR #220, opened 2022-12-08
Fixed login issue

jeffsegal-ccpo:
Made portal page non-public and added a login page such that users are properly redirected to authenticate when accessing /atat


jasonburkert-ccpo:
### Changes
image

image

jasonburkert-ccpo (approved):

PR #221, opened 2022-12-08
Add table Storage Environment Instance

jasonburkert-ccpo:
- Adds table `Storage Environment Instance` (extends `Environment Instance`)

Note: The columns in the base table are sufficient for storage.  Rather than keeping storage instance data in the base table we’ll keep it here.

Ticket: AT-8467


jasonburkert-ccpo:
### Diagram update

image

In wider context
image

juliusfitzhugh-ccpo (approved):

PR #235, opened 2022-12-17
Add Contracting Shop column to Acquisition Package

jasonburkert-ccpo:
- Adds `Contracting Shop` column to `Acquisition Package`

Ticket: AT-8528


jasonburkert-ccpo:
### New column

image

### `Contracting Shop` values

image

### Internal form sample

image

tonysumner-ccpo (approved):

PR #237, opened 2022-12-19
Bugfix/acl_fixes

jeffsegal-ccpo:
After considerable trial and error, we determined that we need to add a create `*` ACL for each Table. Otherwise, unanticipated users will receive 201 responses for POST invocations, but none of their actual fields will be saved (they'll end up with an empty record instead). These fixes allow those POST requests to go through with fields intact.


juliusfitzhugh-ccpo (approved):
From what I can tell, an ACL create has been added to tables that are used for DAPPS with a create `*` ACL.

PR #240, opened 2022-12-20
Make Script Include 'convertToBoolean' more defensive and log debug info

jasonburkert-ccpo:
Defensive and debug logging improvement to script include `convertToBoolean`

- Typical case expects string of value _TRUE/FALSE/YES/NO_ in any case; returns `true` or `false`
- Now logs detailed debug message if string is not an expected value; returns `null`
- Now logs detailed debug message if parameter is undefined/null/object; returns `null`
- Now handles boolean; returns `true` or `false`


jasonburkert-ccpo:
### New implementation

```
function convertToBoolean(input) {
    switch (typeof(input)) {
        case 'boolean': //nothing to do
            return input;
        case 'string':
            var lowercase = input.toLowerCase();
            if (lowercase === "yes" || lowercase === "true") return true;
            if (lowercase === "no" || lowercase === "false") return false;
            gs.debug('Function convertToBoolean() received a string parameter with unexpected value \'' + input + '\'');
            return null;
        default:
            var receivedType = '';
            if (typeof(input) === 'object') {
                if (input === null) {
                    receivedType = 'null';
                } else {
                    receivedType = input.constructor.name;
                }
            }
            if (input === undefined) {
                receivedType = 'undefined';
            }
            gs.debug('Function convertToBoolean() expects a string or boolean parameter.  Received ' + receivedType + ' instead.');
            return null;
    }
}
```

### Old implementation

```
function convertToBoolean(string) {
	var choice = string.toLowerCase();
	if (choice === "yes" || choice === "true") {
		return true;
	}
	if (choice === "no" || choice === "false") {
		return false;
	}
	return null;
}
```

zachclark-ccpo (approved):

jasonburkert-ccpo:
### Sample debug messages
`Function convertToBoolean() received a string parameter with unexpected value ''`
`Function convertToBoolean() received a string parameter with unexpected value 'foobar'`
`Function convertToBoolean() expects a string or boolean parameter.  Received null instead.`
`Function convertToBoolean() expects a string or boolean parameter.  Received undefined instead.`
`Function convertToBoolean() expects a string or boolean parameter.  Received Date instead.`

PR #246, opened 2022-12-22
Update UI Page to use static 'app-js' and 'vendor-js' filenames

jasonburkert-ccpo:
- Updates UI Page to use static `app-js` and `vendor-js` filenames


jasonburkert-ccpo:
### Change to UI Page named _index_

image

juliusfitzhugh-ccpo (approved):
Looks good if only the hashed value was removed. There appears to be coordination with the UI repo, so I'm not merging.

jeffsegal-ccpo (approved):

PR #272, opened 2023-01-11
Update GitHub actions to run on Node 16

jasonburkert-ccpo:
- Updates GitHub actions to use newer versions that run on Node 16.  [Older actions that run on Node 12 are deprecated.](https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/).

Resolves this warning on PR check executions
image

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -17,11 +17,11 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout code
-        uses: actions/checkout@v2
+        uses: actions/checkout@v3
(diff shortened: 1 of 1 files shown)


jasonburkert-ccpo:
### PR check for #1089 with warning
image


### PR check for #1090 without warning
image

jasonburkert-ccpo:
Checked the other repos' use of GitHub actions, this appears to be the only use of deprecated actions at the moment.

juliusfitzhugh-ccpo (approved):

PR #274, opened 2023-01-11
 Add `J&A and MRR Filenames` column to the `Acquisition Package` table

jasonburkert-ccpo:
- Adds `J&A and MRR Filenames` column to the `Acquisition Package` table
- Adjusts form layout

We're going to use this to keep track of the file or files being uploaded and attached to the Acquisition Package record, to make them replaceable but also downloadable with other package documents.

Ticket: AT-8636


jasonburkert-ccpo:
### New column
image

### Internal form
image

PR #277, opened 2023-01-18
Add cloud_support_environment_instance column to Training Estimate table

zachclark-ccpo:
# Overview
Adds new column `Cloud Support Environment Instance` of type _Reference_ to `Training Estimate` table

This will link the Cloud Support Environment Instance to Training Estimate, so we can determine which instance the Training Estimate is estimating.

image

image


jasonburkert-ccpo:
### Diagram update

image

PR #278, opened 2023-01-18
Add 'Instance Number' column to Environment Instance table

jasonburkert-ccpo:
- Adds 'Instance Number' column to `Environment Instance` table

UI relies upon maintained instance numbers internally for order, etc, but these aren't written to database.  To properly generate DoW Task Numbers using the logic given in the ticket we'll need these task numbers to be persisted to the database.  Then they can be used for document generation.

Ticket: AT-8683


jasonburkert-ccpo:
### New column
image

### Adjusted form
All tables that extend `Environment Instance` have a similar form that needed to be adjusted similarly
image

jasonburkert-ccpo:
### Diagram update
image

f600233d1b154d507b782f84604bcb12/dictionary/x_g_dis_atat_environment_instance.xml
```diff
@@ -13,15 +13,16 @@
...
+        <element active="false" label="IGCE Description" max_length="1000" name="igce_description" type="string"/>
+        <element active="false" label="IGCE Title" max_length="200" name="igce_title" type="string"/>
         <element choice="1" label="Instance Location" max_length="40" name="instance_location" type="choice">
             <choice>
                 <element inactive_on_update="false" label="Cloud" sequence="1" value="CLOUD"/>
                 <element inactive_on_update="false" label="On-premise" sequence="2" value="ON_PREM"/>
             </choice>
         </element>
         <element label="Instance Name" max_length="100" name="instance_name" type="string"/>
+        <element label="Instance Number" mandatory="true" max_length="40" name="instance_number" type="integer"/>
```
juliusfitzhugh-ccpo, line 25: I think `Instance Number` works and makes sense, but want to caution that we use the word `Instance` in a lot of different places (but it kind of goes with how things have been build). Also, I'm not sure this captures all the services. 
- `XaaS Service` 4.2.x.x(.x)
  - compute, database, storage, general
  - selected service offerings (this does not extend `EnvironmentInstance`)
- `CDS` 4.2.6 (this does not extend `EnvironmentInstance`)
- `Cloud Support Packages` 4.3.x.x(.x)

Do we want to have one generic name to cover all services, or different naming based on if it is stored as an Instance or not?

To be clear i'm not saying `Instance Number` can't be used, only that we want to make sure we are good with that naming since it will be used in various places and for different services that don't really align with an instance.

Possible alternative names:
- `Task Number`
- `Instance Task Number`
- `Service Task Number`

juliusfitzhugh-ccpo: Didn't have context, ignore the above. 😆

juliusfitzhugh-ccpo (approved):

f600233d1b154d507b782f84604bcb12/dictionary/x_g_dis_atat_environment_instance.xml
jasonburkert-ccpo: This column is expected to contain a simple number.  When used for DoW Task Number it will supply this part of the number -> 4.2.x.x`(.x)`

No matter the rest of the DoW Task Number solution, this seems like a basic piece of info the front-end has today that the back-end does not.  This will help us be consistent.

PR #279, opened 2023-01-19
Fix bugs when generating payload in DOW script include

juliusfitzhugh-ccpo:
Update `Dow` script include based on bugs found from testing with packages created in the UI.
These bugs prevented a Dow payload from being generated and sent to the HOTH API.

### AtatHelper
- Create `recordNotNull` method

### Dow
- Filter null records using `recordNotNull` when calling `getListRecords`. This prevents downstream
logic from breaking due to the object being null.
- Update CDS logic to provide a default `need_for_entire_task_order_duration` when both entire duration 
and selected periods are null/undefined.
- Update PoP `time_frame` to be `null` if not selected.

Ticket: AT-8381


f600233d1b154d507b782f84604bcb12/update/sys_script_include_3d46262b97a751106fa8b4b3f153af90.xml
```diff
@@ -205,6 +205,7 @@ Dow.prototype = {
         awardIds,
         this.AWARDS_HIS_FIELDS,
         this.AWARDS_HIS_TABLE)
+        .filter(Dow.prototype.atatUtil.recordNotNull)
```
jasonburkert-ccpo, line 208: Neat

f600233d1b154d507b782f84604bcb12/update/sys_script_include_3d46262b97a751106fa8b4b3f153af90.xml
```diff
@@ -449,6 +459,12 @@ Dow.prototype = {
           cds["need_for_entire_task_order_duration"] = ctb(cds.need_for_entire_task_order_duration);
           cds["traffic_per_domain_pair"] = JSON.parse(cds.traffic_per_domain_pair);
           if (!cds.need_for_entire_task_order_duration) {
+          // an error occurs when both the entire duration and selected periods is null
+          if (!cds.selected_periods) {
+            // default to entire duration when values are both null 
+            cds.need_for_entire_task_order_duration = true;
```
jasonburkert-ccpo, line 465: This seems to be a reasonable default.  Was this behavior prescribed somewhere?  Just curious.

jasonburkert-ccpo (approved):
Looks good.

f600233d1b154d507b782f84604bcb12/update/sys_script_include_3d46262b97a751106fa8b4b3f153af90.xml
juliusfitzhugh-ccpo: @juliusfitzhugh-ccpo No, just made a judgement call. Ideally, if `CDS` is not required, the value is not used at all. Not sure, about the other place where the same logic is used (I think there is one other place in the script).

jasonburkert-ccpo: Thank you

PR #280, opened 2023-01-19
Make Instance Number column on Environment Instance table not mandatory

jasonburkert-ccpo:
Tweak to #278.  This will avoid integration problems until this field is being populated.

- Makes `Instance Number` column on `Environment Instance` table not mandatory


jasonburkert-ccpo:
### Change to column
image

jeffsegal-ccpo (approved):

PR #285, opened 2023-01-24
Set evaluation Method depending on Source Selection

jasonburkert-ccpo:
- Adds two business rules to set evaluation Method depending on value of Source Selection

Enforces combinations of Source Selection and evaluation Method established by the Evaluation Plan document.

`LPTA` evaluation method is required when Source Selection is `NO_TECH_PROPOSAL`
image

no evaluation method is used when Source Selection is `EQUAL_SET_LUMP_SUM`
image

Ticket: AT-8610


jasonburkert-ccpo:
### First business rule
image

image

jasonburkert-ccpo:
### Second business rule

image

image

juliusfitzhugh-ccpo (approved):
Business rules update the `method` as expected based on the `sourceSelection`.

PR #293, opened 2023-01-31
Environment schema changes for provisioning a Portfolio

juliusfitzhugh-ccpo:
Change in provisioning schema to include *Environments* which are children of a *Portfolio*.
### Create `Environment` table
CreateEnvironmentTable_AT-8690
### Update `Portfolio` table
- Add column `Environment` - ref
AddEnvironmentPortfolio_AT-8690
- Make columns inactive
InactiveColumnsPortfolio_AT8690
### Update `Provisioning Job` table
- Add column `Environment` - ref 
AddEnvironmentColumnProvisioningJob_AT-8690
- Make `Acquisition Package` column inactive
InactiveColumnsProvisioningJob_AT-8690

### Update `Operator` table
- Make `Portfolio` column inactive
InactiveColumnsOperator_AT-8690

### Other
- Update diagram based on schema changes

Ticket: AT-8690


zachclark-ccpo:
Is 60 characters long enough for the Name field?

juliusfitzhugh-ccpo:
> Is 60 characters long enough for the Name field?

I was wondering the same. I left it the same since it was previously used and I believe based on business rules we had previously.

jasonburkert-ccpo:
> > Is 60 characters long enough for the Name field?
> 
> I was wondering the same. I left it the same since it was previously used and I believe based on business rules we had previously.

Looked into this.  We have a 60 char limit for Acquisition Package.Project Title, so a 60 char limit for this still seems reasonable given no other direction.

zachclark-ccpo (approved):

juliusfitzhugh-ccpo:
Given the green light that 60 should be good for the `name`.

jasonburkert-ccpo:
Forgot to update the diagram.

PR #296, opened 2023-02-05
Fix IGCE DocGen SNOW side bugs #1

zachclark-ccpo:
# Bugs fixed
## Travel Contract Type:
image

- By default, all travel instances should have “T&M” for Contract

## Quantity Unit mapping:
image
Reference package 454

Option year 1 has duration of 1 Year

In IGCE, it’s displaying it as “1 month”

image
Reference package 455

Option year 2 has duration of 300 Days

In IGCE, it’s displaying it as “300 month”

### Fix
Now uses the Unit Quantity sys_id value pairs exclusively for the Period and units, taking the Units from sys_id Period reference instead of relying on the Unit field of the IGCE Estimate

Ticket: [AT-8736](https://ccpo.atlassian.net/browse/AT-8736)

[AT-8736]: https://ccpo.atlassian.net/browse/AT-8736?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


juliusfitzhugh-ccpo (approved):
Looks, good. I re-ran package 455 and the unit updated so that the quantity and unit is `300 days`. No travel items in package 455, but the code has `T&M` as the contract type for travel.

PR #302, opened 2023-02-10
Add column Contract Type.Display

jasonburkert-ccpo:
- Adds calculated value column `Display` to the `Contract Type` table
  - Its value is based on two boolean columns in the table
  - This is now the ServiceNow _Display_ column for that table
- Adjusts form and list layouts to include new column
- Changes existing column `Contract Type` on `IGCE Estimate` to be a calculated value based on new column
  - Values of new column match values of this existing Choice column ("_FFP_/_T&M_/_TBD_") so updates are seamless
  - It's read only so it can not be set manually
  - This supports existing implementation and can eventually be removed following refactoring
- Updates diagrams.md to reflect new change
- Sets label for Portfolio table - label `ATAT:Portfolio` was lost recently, this is an attempt to fix it

Ticket: AT-8753


jasonburkert-ccpo:
### Changed existing column to be calculated
image

### Form view

image


### List view

image

juliusfitzhugh-ccpo (approved):
Looks good. The `Contract Type` display the correct label based on if FFP, T&M, or both are selected.

PR #303, opened 2023-02-14
Update `Send Package to DITCO` flow

juliusfitzhugh-ccpo:
- Update the subject, message, and CC of the email sent to DITCO inside the 
Send Email action of the Send Package to DITCO flow.

Ticket: AT-8722


jasonburkert-ccpo (approved):
Looks good.  Examined changes and tested in a PDI.
Suggested one minor change to email subject in ticket.

PR #304, opened 2023-02-14
IGCE Portability Plan Mapping and currentEnvironment Bug Fixes

zachclark-ccpo:
## Portability Plan
Now properly using `item.title` instead of `item.serviceTitle` for the if statement mapping, also added a null check on `item.title`.
Test on `ACQ0000469`, should only map to Base Period.

## Current Environment
Now checking with `.toString()` due to inconsistency with SNOW data types.
Should properly map Instructions Page options (test on `ACQ0000455` for currentEnvironment exists)

## Travel Options
Added null check for igceInstructions to prevent improper quantity mapping if instructions don't exist.

https://ccpo.atlassian.net/jira/software/c/projects/AT/boards/47?modal=detail&selectedIssue=AT-8735
https://ccpo.atlassian.net/browse/AT-8736


juliusfitzhugh-ccpo (approved):
Other than fixing merge conflict, looks good. Tested with package 469 for portability plan.

PR #306, opened 2023-02-16
Update values for `Status` column of `Provisioning Job` table

jasonburkert-ccpo:
- Updates values for `Status` column of `Provisioning Job` table
  - removed _FAILED_
  - added _FAILURE_
  - removed _COMPLETE_
  - added _SUCCESS_
- Updates step 1 of action `PROV: Consume Queue Payload` in which a script updates job status values

Another use of a job status value was examined, but required no changes.
Flow `PROV: Provisioning Request` uses a value of _FAILURE_ in step 7.

Ticket: AT-8780


juliusfitzhugh-ccpo (approved):
Looks good. Nothing stood out and it appears the correct locations were updated.

PR #313, opened 2023-02-24
Create Unsigned and Signed records for an Acquisition Package

juliusfitzhugh-ccpo:
Add logic to create an unsigned and signed record for a package if the
records do not already exist. This ensures the records exist for attaching
documents created or uploaded after document generation is completed.
### Create a record that reference the package in the following tables if one does not exist
- `DAPPS:Package Documents Unsigned`
- `DAPPS:Package Documents Signed`

It was considered to place the creation of the records in the
`DOCGEN: Prepare Acquisition Package Documents` flow or in the
`UTIL: Copy Package Attachments for Download` action. However,
most of the logic interacting with the two tables was provided in the
initialize method of the PackageAttachments script include, which was
already looking up the records if they exist.

The other locations would require adding what seemed like repeated logic to look
up the records and then create records if no record was found.

Ticket: AT-8797


jeffsegal-ccpo (approved):
Hey it's code I can actually review in a PR, how about that? Looks good.

PR #314, opened 2023-02-24
Fixed various problems introduced in AT-8737

jeffsegal-ccpo:
Fixed several typos and assorted problems in `Igce` related to the introduction of support for Architectural Design Requirements in AT-8737.


jasonburkert-ccpo (commented):
This looks good upon initial review.  Going to try it via background script in PDI.

juliusfitzhugh-ccpo (approved):
Ran packages 105, 111, 112, and 150 through the `Igce.getPayload()` and also in subflow for generating IGCE.  Things appeared to working as expected. 

Looks like there is some clean up needed in the description column for all caps, but can be addressed in a different ticket. 
IGCE-DescriptionColumn

PR #319, opened 2023-03-09
Make service portal components more accessible to ServiceNow users

jasonburkert-ccpo:
## Background

ServiceNow includes a feature called [Service Portal](https://www.servicenow.com/products/service-portal.html) of which DAPPS/ATAT make limited use.
- We use a single _Service Portal Page_ `atat_portal_home_page` to provide CSS
- We use a single _Service Portal Widget_ `atat_app_iframe` to provide an iframe whose `src` is the endpoint for a _UI Page_ named `index`
- See [this comment](https://github.com/dod-ccpo/atat-snow/pull/319#issuecomment-1463922597) for more detail

image

### Where to find these components in Studio
image

## Changes
This PR makes that page and widget more accessible so that more ServiceNow users can access the ATAT scoped application.

- Adds _snc_external_ and _snc_internal_ roles to ACLs for ATAT App iframe widget and ATAT Portal Home Page


jasonburkert-ccpo:
## The _Service Portal Page_ w/ CSS

image

## The _Service Portal Widget_ w/ iframe

image

## The _UI Page_ whose endpoint is referenced by the iframe

image

PR #318, opened 2023-03-08
Add new column and two service offerings

jasonburkert-ccpo:
### This new column will be used to store the Y/N selection for whether design services are needed
- Adds column `needs_architectural_design_services` to `Architectural Design Requirement` table

### This new group will contain two entries representing no services, one for offering type _XaaS service_ and one for offering type _Cloud support_
- Adds service offering group `NONE_APPLY`

### Two new entries in reference table `Service Offering` will be used to represent no services and belong to the new group mentioned above
- Adds two new reference data rows to `Service Offering` table

PR #320, opened 2023-03-10
Set sequence number for new service offering group NONE_APPLY

jasonburkert-ccpo:
- Sets sequence number to 99 for service offering group `NONE_APPLY`.  This was missing from #318 
- Sets sequence numbers to 99 for two new service offerings for _None of these apply to my acquisition_


jasonburkert-ccpo:
### New sequence number

This had no value before making this PR necessary
image

### Updated sequence numbers

These were good enough, but update to 99 for congruence with the above.  These options will always be at the bottom of lists.
image

jasonburkert-ccpo:
OBE.  Discussed overlapping change with @ericyoungquist-ccpo

PR #323, opened 2023-03-15
AT-8871: Convenience data for UI

jeffsegal-ccpo:
The UI needs access to a given Portfolio's Vendor and a given Environment's Classification. It's much easier to simply make these available in the back end than forcing the UI to navigate the data structures and perform multiple queries to obtain the info, so we implemented it here.

Specifically:

- Adds `Portfolio.Vendor` column
- Adds `Environment.Classification Level` column
- Adds logic to set them on initial provisioning (in `Provisioning/Start Provisioning` resource in the `Provisioning` scripted REST API)


jasonburkert-ccpo:
### New column
image

In action on internal form
image

### New column

image

In action on internal form
image

### Updated Scripted REST resource

Sets the new columns when provisioning starts
image

jasonburkert-ccpo (approved):

PR #325, opened 2023-03-17
AT-8769: Updates to support consuming Create Portfolio responses

jeffsegal-ccpo:
- Replaced `Environment` column with `Portfolio` column in `Provisioning Job` table
- Updated flow `PROV: Queue Initial Portfolio Provisioning Job` to set `Provisioned Request Date` again
- Removed `Dashboard Link` column from `Portfolio` table as it only pertains to `Environment` now
- Updated action `PROV: Consume Queue Payload` to handle success and failure responses for ADD_PORTFOLIO operation


f600233d1b154d507b782f84604bcb12/update/sys_hub_flow_a0ff1bf42fd15110e58359a72799b63b.xml
```diff
@@ -10,13 +10,13 @@
...
         <master_snapshot>99d63b37dbb619d0b1227ea5f396190e</master_snapshot>
         <name>PROV: Queue Initial Portfolio Provisioning Job</name>
```
jasonburkert-ccpo, line 15: May need to remove or alter this condition in the trigger of flow `PROV: Queue Initial Portfolio Provisioning Job`

image

jasonburkert-ccpo:
### New column
Looks good.  I wonder if the old column should remain and made inactive instead.
image

### Date set in flow
@jeffsegal-ccpo  Is this the date that was updated?  I'm not sure I have the right place because it doesn't match the description.  But also, this script is empty.

image

### Updated action
image

jasonburkert-ccpo (commented):
See my question about the date set in the flow.

jasonburkert-ccpo:
Highlighted issues will be addressed in a follow on PR as discussed per chat

jasonburkert-ccpo (approved):

PR #333, opened 2023-03-24
AT-8821: Small fixes

jeffsegal-ccpo:
- Made JWCC contractor validation case-insensitive
- Added support for sending cloud distinguishers in create environment payloads


jasonburkert-ccpo:
Note for future improvement.

Contractors extracted from EDA should be uppercased immediately to agree with Vendor choices in ServiceNow.  Currently they are transformed on use in downstream code.

image

jasonburkert-ccpo (approved):

jasonburkert-ccpo:
Note for future improvement.

Contractors extracted from EDA should be uppercased immediately to agree with Vendor choices in ServiceNow.  Currently they are transformed on use in downstream code.

image

PR #348, opened 2023-04-11
bugfix/AT-8537-platform-DISA-org-dropdown-not-populating

tonysumner-ccpo:
changed 2 business rules so that DISA_organization would save successfully in org tbl


jeffsegal-ccpo (approved):

jasonburkert-ccpo:
Which two business rules were changed?  What specifically changed?

PR #349, opened 2023-04-12
AT-8848: Added CONTRIB: Send Email Upon Invitation Flow

jeffsegal-ccpo:
Added a new Flow which sends emails to all newly added Contributors for any given Acquisition Package. 

image

image


jasonburkert-ccpo (changes requested):
Recent fixes in response to identified issues look good.  Think there is one more update needed for package title display in the body.

jeffsegal-ccpo:
> Recent fixes in response to identified issues look good. Think there is one more update needed for package title display in the body. I'm testing a fix.
> 
> image

Ahh good catch. I swear that was correct at some point.

jasonburkert-ccpo (approved):

PR #351, opened 2023-04-13
Add flow to notify new package owner of ownership transfer

jeffsegal-ccpo:
Implemented Flow `CONTRIB: Notify New Package Owner`, which sends email notifications to users who have been added as Mission Owners for DAPPS Acquisition Packages. Uses a Subflow which is shared between it and the Flow responsible for notifying Contributors (`CONTRIB: Send Email Upon Invitation`). The emails themselves are sent in custom Actions for each Flow.

Note: I also refactored `CONTRIB: Send Email Upon Invitation` to re-use these constructs.  

# Flows

## Contributors
image

## Mission Owners
image

# Shared Subflow
image

# Actions

## Contributors
image

## Mission Owners
image

Ticket: feature/AT-8765


jasonburkert-ccpo (approved):
The new Flow, updated Flow, new Subflow, and two new Actions look good.

As a flow name `CONTRIB: Notify New Package Owner` seems mildly misleading, but I think that's outweighed by the ability to filter on a common prefix for flow components related to these notifications.

PR #354, opened 2023-04-18
PR 354 - Feature/AT-8953-address-table

DGabriel-CCPO:
New table `Address Table` was added.  Existing table `Acquisition Package` has a new column `contracting_shop_non_ditco_address` that references the new table.

Ticket: AT-8953


jasonburkert-ccpo:
I left some feedback in Slack

jasonburkert-ccpo (changes requested):
The improvements to the country column, and the new Choice type columns look great.

For some reason the `Address Type` field on the internal form is behaving like a drop-down instead of a simple textfield entry for a string.  Suggest recreating it.  If that fails to work, recreate it using a different column name (apart from `address_type`).

image

f600233d1b154d507b782f84604bcb12/dictionary/x_g_dis_atat_acquisition_package.xml
```diff
@@ -1,6 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?><database>
-    <element label="DAPPS:Acquisition Package" max_length="40" name="x_g_dis_atat_acquisition_package" sizeclass="163" type="collection">
+    <element db_object_id="3a80dd5e971121107b1372571153af2b" label="DAPPS:Acquisition Package" max_length="40" name="x_g_dis_atat_acquisition_package" type="collection">
         <element label="ACOR" max_length="32" name="acor" reference="x_g_dis_atat_contacts" type="reference"/>
+        <element label="Address" max_length="32" name="address" reference="x_g_dis_atat_dapps_address_table" type="reference"/>
```
jasonburkert-ccpo: It's not clear to me what this new column `Address` in `Acquisition Package` table is for.  New table `Address Table` has a reference to `Acquisition Package`, so _many_ addresses can be associated with a single package.  But this column would be used to associate a _single_ address to a package; this appears to be redundant.

jasonburkert-ccpo (changes requested):
New table and the fix to Address Type looks good.  See my comment re: the change to Acquisition Package

image

jasonburkert-ccpo (approved):
Much clearer now!  Thanks

jasonburkert-ccpo:
## New table `Address Table`
image

## New column in `Acquisition Package` referencing the new table
image

PR #373, opened 2023-05-08
PR 373 - updated current contract information to add the following:

tomarnold-ccpo:
* Acquisition Package reference
* Business Size
* Competitive Status
* Start Date


f600233d1b154d507b782f84604bcb12/author_elective_update/sys_choice_x_g_dis_atat_current_contract_and_recurring_information_current_contract_exists.xml
```diff
@@ -0,0 +1,79 @@
+<?xml version="1.0" encoding="UTF-8"?><record_update>
+    <sys_choice action="INSERT_OR_UPDATE" field="current_contract_exists" table="x_g_dis_atat_current_contract_and_recurring_information" version="3">
+        <sys_choice_set action="INSERT_OR_UPDATE">
+            <element>current_contract_exists</element>
+            <name>x_g_dis_atat_current_contract_and_recurring_information</name>
+            <sys_class_name>sys_choice_set</sys_class_name>
+            <sys_created_by>tom.arnold</sys_created_by>
+            <sys_created_on>2023-05-07 03:50:18</sys_created_on>
+            <sys_mod_count>0</sys_mod_count>
+            <sys_name>current_contract_exists</sys_name>
```
jasonburkert-ccpo, line 10: What's the meaning of this column now that a Package can have many prior contract records?  It seems like having this in this table opens the possibility for conflicting information among child records for the package.  I think this might belong in the Acquisition Package table now.

Because this is essentially UI state for the package, to my knowledge, and now there is more than one place to store that state for a single package.

jasonburkert-ccpo (changes requested):

f600233d1b154d507b782f84604bcb12/author_elective_update/sys_choice_x_g_dis_atat_current_contract_and_recurring_information_current_contract_exists.xml
jasonburkert-ccpo: I started having this thought in the ticket, and now forcing the discussion.  @ericyoungquist-ccpo @tonysumner-ccpo @jeffsegal-ccpo Thoughts?

jeffsegal-ccpo: My guess is that the UI needs to be updated to query the Current Contract table to determine this rather than rely on this boolean.

ericyoungquist-ccpo: UI still needs to know what user checked for this radio group. Field name could be updated to `current_or_previous_contract_exists` possibly to be more accurate of what it is
image

jeffsegal-ccpo: That makes sense. I'd vote just to leave it alone.

jasonburkert-ccpo (approved):

PR #383, opened 2023-05-17
Update action `MIGRATION: Contract History`

jasonburkert-ccpo:
- Activates column `Acquisition Package.Current Contract Information` which is necessary to execute the migration action
- Updates action `MIGRATION: Contract History`
  - Requires these columns:
    - `x_g_dis_atat_acquisition_package.current_contract_and_recurring_information` - identifies the child record before migration - must be non-null
    - `x_g_dis_atat_current_contract_and_recurring_information.acquisition_package` - will identify the parent record after migration - must be empty
  - Stops if no records to migrate
  - Logs details to application log

Ticket: AT-9051


jasonburkert-ccpo:
### Action steps

Added condition - effectively skips records that already contain a package in the target column.  The action can be executed multiple times and eventually there should be zero records updated.
image

Log a clear message about records to be updated - this count was always zero when column `Acquisition Package.Current Contract Information` was inactive
image

The core of this is the same, but there is now additional application logging
image

Log a clear message about records that were updated
image

jasonburkert-ccpo:
### Execution examples

Two acquisition packages being updated.
image
Application log contains more details
image

When there are no acquisition packages [left] to update
image

jeffsegal-ccpo (approved):
Appreciate all the detail!

PR #402, opened 2023-06-08
Improve Cost Estimate REST API; Add External Ordering Agency Fee, etc.

jasonburkert-ccpo:
- Adds property `surge_and_fees`; replaces calculation code comments with `Description` object property that contains calculation details within the relevant parts of the JSON response

Ticket: AT-9008


jasonburkert-ccpo:
I changed so much, just going to replace this.

PR #406, opened 2023-06-15
AT-9095: Updated POST /provisioning API

jeffsegal-ccpo:
Modified API to accept cspName rather than inferring CSP from vendor and classification level

Ticket: AT-9095


jasonburkert-ccpo (approved):
Looks good.

I don't recall seeing that `$DISPLAY` syntax before.

PR #407, opened 2023-06-16
Add contextual comment re: IDIQ Clins to the EDA script include

jasonburkert-ccpo:
Think this context is helpful after a discussion with @ericyoungquist-ccpo

image


jasonburkert-ccpo:
### Sourced from IGCE template reference sheet
image

ericyoungquist-ccpo (approved):

PR #424, opened 2023-07-10
AT-9125: Flow to send email upon Portfolio provisioning.

stephenhayes-ccpo:
Ticket description included this reference image.  Guidance from Eric said plain-text is fine, but I already had some HTML in so I've left it as is.  
image

This picture shows the email.
snowEmailLog

And here is the formatted HTML which I have been assured will render correctly in production.
formattedHtml


luakim-ccpo:
@stephenhayes-ccpo don't include watermark
image

luakim-ccpo:
@stephenhayes-ccpo 
image
poc should be the user who provisioned the portfolio - it should be portfolio owner

add "PROV:" prefix in the flow name

stephenhayes-ccpo:
> @stephenhayes-ccpo image poc should be the user who provisioned the portfolio - it should be portfolio owner
> 
> add "PROV:" prefix in the flow name

PROV: added to flow name.
watermark removed
POC in email body switched to "updated_by".

luakim-ccpo (approved):

PR #425, opened 2023-07-11
Feature/at 9061

torinharthcock-ccpo:
Updated IFP script to use the costEstimate function to calculate base year subtotal.


torinharthcock-ccpo:
image

image

luakim-ccpo:
@torinharthcock-ccpo  This only grab the value for DITCO user. if the user is non-ditco estimated base year value should be Subtotal for base period (no Surge) PLUS External Ordering Agency Fee (1%)

torinharthcock-ccpo:
> @torinharthcock-ccpo This only grab the value for DITCO user. if the user is non-ditco estimated base year value should be Subtotal for base period (no Surge) PLUS External Ordering Agency Fee (1%)

According to Devonte, the costEstimate function calculates the subtotal as including the external Ordering Agency fee. There's been some issues with the costEstimate function for non-Ditco users but my understanding from Tom is that that would be addressed in a separate ticket.

luakim-ccpo:
Documenting comment made to Torin in Slack
- DITCO user - subtotal of all CLINS for base year without surge
- Non-DITCO user - (subtotal of all CLINS for base year without surge) * 1.01 [adding 1% of subtotal]

With this, you are keeping your base year subtotal for DITCO user. 
However, for non-ditco user, you have to multiply by 1% to include external agency fee. 
you are grabbing the right data from cost estimate api, you just need extra logic for non-ditco user

luakim-ccpo:
image

luakim-ccpo (approved):

PR #427, opened 2023-07-11
AT-9337: Working implementation for emailing new portfolio viewers.

stephenhayes-ccpo:
Testing done by manually adding and removing viewer in the portfolio table.

Capture
Capture2


luakim-ccpo:
@stephenhayes-ccpo 
1. Invalid flow name . "PROV:" is too generic and doesn't give insights to what this flow is for
2. Body only contains text "test"
image
3. You need to check when viewer is added. I would strongly recommend to reference CONTRIB: Send Email Upon Invitation flow.

stephenhayes-ccpo:
> @stephenhayes-ccpo
> 
> 1. Invalid flow name . "PROV:" is too generic and doesn't give insights to what this flow is for
> 2. Body only contains text "test"
>    image
> 3. You need to check when viewer is added. I would strongly recommend to reference CONTRIB: Send Email Upon Invitation flow.

Oops! I think I somehow stashed my implementation prior to committing (???)
Good catch on the name.

luakim-ccpo:
image

This is what i see in the HTML viewer
image

luakim-ccpo (approved):

PR #435, opened 2023-07-20
feature/AT-9427-Add-Portfolio-Viewers-And-Managers

stephenhayes-ccpo:
I split the original FLOW into two; one for manager's and one for viewers.  Manager FLOW triggers off a change to managers; same for viewers.  Did this because the email recipients get a different email body and I didn't see a decent way to do this on one flow.


AndrewNickerl-ccpo (approved):

stephenhayes-ccpo:
Previous flow split into two.

image

For example, the flow for Viewers responds with viewer-specific copy:

image

ericyoungquist-ccpo (commented):
approved

PR #438, opened 2023-07-21
AT-9333 Adding in business rule to trigger emails when ownership is changed

loganjohnston-ccpo:
We will be utilizing the search feature in Servicenow for moving around to these changes: 
image
### Business Rule: 
Search 'Business Rule' and select the one under System Definition.

image
This should bring up a table which also has a search feature and in this search we will look for 'Portfolio Ownership Change' Currently there is only one record here.

image
Once we click into this we can see the Name and Table (portfolio table) fields and tabs further down the page 'When to run', 'Actions', and 'Advanced'

_When to run_: 
image
This tab shows us the condition in which to run our business rule, in this case when Portfolio Owner changes.

Not using the 'Actions' Tab.

_Advanced_:
image
This tab shows us the function that is fired off when the business rule 'When to run' is met. 
This function grabs the current portfolio owners ID which we use to capture the current portfolio owners email and pass it along as a parameter to our notification via the eventQueue. 

### Event:
Search 'Events' and select 'Registry' under System Policy.

image
This should bring up a table which also has a search feature and in this search we will look for 'Portfolio' 
Currently there is only one record here.

image
Once inside there is just some text defining and describing the event we created.

image

### Notification:
Search 'Notification' and select the one under System Notifications => Email

image
This should bring up a table which also has a search feature and in this search we will look for 'Portfolio Ownership Change' Currently there is only one record here.

image
Once inside we will see some more tabs 'When to send', 'Who will receive', and 'What it will contain'

_When to send_:
image
the 'Send When' drop down should be 'Event is fired' and you should see the event from the previous step in the 'Event name' dropdown

_Who will receive_:
image
The email that we passed via the evenQueue param1 should be used as the recipient

_What it will contain_:
image
Should see both html and plain text content types, omitting the watermark (this shows up as REF:#xxxx on the email if we don't, referencing the event #) and under Email Template we should see the name of our template 'Portfolio Ownership Change' 

### Template:
Search 'Templates' and select the one under System Notifications => Email

image
This should bring up a table which also has a search feature and in this search we will look for 'Portfolio Ownership Change' There are multiple records here but only one with this name exactly.

image
Once inside you will see an html email template plus a section for just the text

### Email script: 
Search 'Notification Email Scripts', there should only be one result. 

image
This should bring up a table which also has a search feature and in this search we will look for 'get_updated_by_name'
There should only be one record.

image
Once inside we should see a runMailScript function that gets the sys_updated_by user name and triggers a template print to return the user name. This is used in the template to get the previous owners user name


torinharthcock-ccpo:
Is there an easy way we can change the font on these? 
image

torinharthcock-ccpo (approved):

PR #441, opened 2023-07-25
Remove notification/unsubscribe from email footer

AndrewNickerl-ccpo:


loganjohnston-ccpo (approved):
Looks good, watermark and unsubscribe text was removed

PR #448, opened 2023-08-01
AT-9496 - Adjusted "Request" script include and "UTIL: Make a Request" action.

xanderkeele-ccpo:
Adjusted "Request" script include to allow for no mutualAuth and "UTIL: Make a Request" action to handle a credential id.


f600233d1b154d507b782f84604bcb12/update/sys_hub_action_type_definition_738c03a4979211106fa8b4b3f153af63.xml
```diff
@@ -150,14 +150,26 @@
...
  
+  var token = inputs.token
+  if (inputs.credential_id) {
+    try {
+      var provider = new sn_cc.StandardCredentialsProvider();
+      var credential = provider.getCredentialByID("x_g_dis_atat.credentials." + inputs.credential_id);
+      if (credential) {
+	      token = credential.getAttribute("password")
+      }
+    } catch (error) {
```
jeffsegal-ccpo, line 171: We should at least log a warning if the given `credential_id` could not be found, particularly since we are doing some additional logic that folks might not be expecting.

f600233d1b154d507b782f84604bcb12/update/sys_script_include_5d76f38e970611106fa8b4b3f153afc4.xml
```diff
@@ -32,7 +32,9 @@ Request.prototype = {
 		request.setEndpoint(req.endpoint);
 		request.setHttpMethod(req.method);
 		request.setLogLevel("all");
-		request.setMutualAuth(req.mutualAuth || this.JKS_443);
+		if (req.mutualAuth) {
```
jeffsegal-ccpo: This is reasonable but we also need to make sure that calling Flows were not relying on the default value that was removed (`this.JKS_443`).

jeffsegal-ccpo (changes requested):

f600233d1b154d507b782f84604bcb12/update/sys_script_include_5d76f38e970611106fa8b4b3f153afc4.xml
xanderkeele-ccpo: Yeah, thought of a better way overnight to maintain existing functionality.

f600233d1b154d507b782f84604bcb12/update/sys_hub_action_type_definition_738c03a4979211106fa8b4b3f153af63.xml
xanderkeele-ccpo: It's catches the exception and logs it. What more would you like to see?

PR #462, opened 2023-08-08
hotfix - user requesting provisioning becomes owner AND manager

ericyoungquist-ccpo:
Added user requesting provisioning to portfolio managers in addition to owner


ericyoungquist-ccpo:
was not created off latest `develop` - closing

PR #478, opened 2023-08-16
AT-9613: BUG: Added proper conditionals to fix the bug

xanderkeele-ccpo:
Added code. Need to identify proper test method before completing.

image


xanderkeele-ccpo:
Tested in DSP Dev with @luakim-ccpo

jeffsegal-ccpo (approved):

PR #483, opened 2023-08-21
AT-9639: Developer Navigation Footer & GitHub API Page size

xanderkeele-ccpo:
AT-9639: Added input to `CD: Update UI Files` specifying whether to include Developer Navigation footer.
Added a value of "true" to `CD: Deploy from Develop` subflow.
Added `per_page=250` query param to `CD: Deploy from Develop` for GitHub API call.

Changes to `CD: Update UI Files` text substitution code.
image

Changes to `CD: Deploy from Develop` `per_page` query param
image

Source from DevTools in `DSP Dev` after running `CD: Deploy from Develop`
image

Source from DevTools in `DSP Dev` after running `CD: Deploy From Release`
image


xanderkeele-ccpo:
Interesting observation: 

The following code, found in ATATFooter.vue shows how the source is written.
image

That expands to:
image

So, the fix was to add the VUE_APP_allowDeveloperNavigation environment variable directly in the expanded representation of `proccess.env` in the resulting `app.js` file.

jeffsegal-ccpo (approved):
Appreciate all the detail. Looks good.

PR #484, opened 2023-08-21
feature AT-9101 portfolio last updated

loganjohnston-ccpo:
Added in a business rule to update the associated portfolio last updated date/time to that of the new environment change.
image

This business rule fires off whenever Environment is updated:
image

AC:
1. Date must reflect changes to the overall Environments


xanderkeele-ccpo (approved):
Approved

PR #485, opened 2023-08-21
Feature/AT-9588

jeffsegal-ccpo:
Implemented CRUD ACLs on the Portfolio Table via a new `PortfolioAclService`.

## Create (Delete looks the same):

image

## Read (Write is the same except with a different enum):

image

```js
var AclService = Class.create();
AclService.prototype = {
    initialize: function() {
    },
	
	/**
   * Concatenates an array of comma-delimited roles. E.g., given ["a,b", "c", "d,e"], returns "a,b,c,d,e".
   *
   * @param {string} roleLists - a list of comma-delimited lists of roles
   * @return {string} a single concatenated comma-delimited list of roles
   */	
	constructAccessList: function(roleLists) {	
		let accessList = "";
		if (roleLists) {
			roleLists.forEach(function(roleList) {
				if (roleList) {
					accessList += accessList ? "," : "";
					accessList += roleList ? roleList : "";					
				}
			});
		}
		return accessList;
	},
	
	/**
   * Returns a boolean indicating whether the current user has access to a given resource.
   *
   * @param {object} aclObject - custom object with relevant properties for the ACL decision
   * @return {boolean} authorization decision
   */	
	hasPermission: function(aclObject) {
		var accessList = aclObject.accessList.toString().split(",");
		var userHasPermission = listContains(aclObject.userSysId, accessList);
		var matchedRoles = aclObject.userRoles.filter(function(userRole) {
			return listContains(userRole, aclObject.allowedRoles);
		});
		var userHasRole = matchedRoles.length > 0;
		return userHasPermission || userHasRole;
	},

    type: 'AclService'
};
```

```js
var PortfolioAclService = Class.create();
PortfolioAclService.prototype = Object.extendsObject(AclService, {
	// Note: we don't include CREATE or DELETE because we don't need custom code for those operations (see the Portfolio ACLs themselves for more)
	OPERATIONS: {
		READ: "read",
		WRITE: "write",
	},
	
    initialize: function() {
    },
	
	constructAclObject: function(operation) {		
		let owner = current.portfolio_owner ? current.portfolio_owner.toString() : null;
		let managers = current.portfolio_managers ? current.portfolio_managers.toString() : null;
		let viewers = current.portfolio_viewers ? current.portfolio_viewers.toString() : null;

		const aclService = new AclService();
		let accessList;
		
		if (operation === this.OPERATIONS.READ) {
			accessList = this.constructAccessList([owner, managers, viewers]);			
		}
		else if (operation === this.OPERATIONS.WRITE) {
			accessList = this.constructAccessList([owner, managers]);
		}
		else throw "Unsupported operation: " + operation;

		return {
			userSysId: gs.getUserID(),
			userRoles: gs.getUser().getRoles().toString().split(","), 
			accessList: accessList,
			allowedRoles: ["x_g_dis_atat.hacc_admin"]
		};

	},

    type: 'PortfolioAclService'
});
```


xanderkeele-ccpo (approved):
Approved

PR #486, opened 2023-08-22
AT-9588: Hotfix

jeffsegal-ccpo:
Accounted for the "Explicit Role Plugin", which automatically adds roles to ACLs. That is, Portfolio create and delete ACLs no longer have the automatically added `snc_external` role:

image

And read/write require either `snc_internal` or `snc_external`:

image

This is required because the Explicit Role Plugin automatically adds roles to ACLs which do not have any.


xanderkeele-ccpo (approved):
Approved

PR #490, opened 2023-08-24
fix for business rule to only email operators in unclassified environ…

stephenhayes-ccpo:
**Updated business rule to only email operators in unclassified environments.**

### Application Log: only one email is sent to an operator and only if it's an unclassified environment.
appLogs

### Email Log:
emailLog

### Code
cspemailadmin


xanderkeele-ccpo (approved):
Approved.

PR #491, opened 2023-08-24
Removed checksum.txt and added it to the .gitignore

xanderkeele-ccpo:
Removed the checksum.txt file permanently and added it to .gitignore to ensure it doesn't return. 

Reference https://docs.servicenow.com/bundle/tokyo-application-development/page/build/applications/concept/c_SourceControlIntegration.html to verify it is not needed.


jeffsegal-ccpo (approved):
Approved with extreme prejudice.

PR #503, opened 2023-09-05
feature/AT-9685-update-portfolio

loganjohnston-ccpo:
Task order number was being updated with a javascript array reference. toString() fixed the issue. 
image


xanderkeele-ccpo (approved):
Approved.

PR #506, opened 2023-09-07
AT-9565

stephenhayes-ccpo:
- emails dispatched individually instead of an aggregate


stephenhayes-ccpo:
Changed logic to dispatch individual emails instead of an aggregate.

individualEmails

ericyoungquist-ccpo (approved):

PR #507, opened 2023-09-07
AT-9691: Create Flow to invoke HOTH API TO update operation

xanderkeele-ccpo:
Created new flow to create provisioning job record for new UpdateTaskOrder in CSP. Modeled after `PROV: Queue Initial Portfolio Provisioning Job` and triggered on any record update on the `ATAT:Task Order` table with the condition that it has a `csp_id` field already populated.
image

image

image

`HothProvisioning` Script Include was also updated with a new function to create the UpdateTaskOrderRequest object to send to the HOTH API.
image

Adjusted `PROV: Consume Queue Payload` to account for the new `UPDATE_TASK_ORDER` job type
image

Also added `CSP Id`, `Provisioned`, and `Provisioned Date` fields to `ATAT:Task Order` table.


jeffsegal-ccpo (approved):
Looks perfect and all the detail in the PR is awesome

PR #509, opened 2023-09-08
updated script to include the funds_total when pulling data from eda …

tomarnold-ccpo:
…for new and existing clins
resolveClins() Lines 385-387
image

createCLINGlideRecord() Lines 407-409
image

Added key/value pair Line 464
image


stephenhayes-ccpo (approved):
Line 400 has a call to update() inside a loop; that may result in _n_ DB updates for a collection of _n_ items. I say may because I don't know if that's a synchronous call or if it's async and processed later; it's also out of scope for the ticket. Just something to consider though.

tomarnold-ccpo:
@stephenhayes-ccpo This is a SNOW construct on how to iterate over your records and update them.  For reference, take a look [here](https://docs.servicenow.com/bundle/tokyo-api-reference/page/script/server-scripting/concept/c_UsingGlideRecordToQueryTables.html) and search for `update`

PR #523, opened 2023-09-25
AT-9482: Implemented /user_proxy/self/roles operation

jeffsegal-ccpo:
This operation will allow the UI to easily query to determine if the current user has a given role. Includes a `filter` parameter in case we are concerned about a large number of roles coming back in the response.

```js
(function process(/*RESTAPIRequest*/ request, /*RESTAPIResponse*/ response) {
	const filter = request.queryParams.filter;
	const roles = [];
	const userRole = new GlideRecord("sys_user_has_role");
	userRole.addQuery("user", gs.getUserID());
	userRole.query();
	while (userRole.next()) {
		roles.push(userRole.role.getDisplayValue());
	}
	if (!filter) {
		return roles;
	}
	else {
		const filterRoles = filter.toString().split(",");
		const filtered = roles.filter(function (role) {
			return filterRoles.includes(role);
		});
		response.setStatus(filterRoles.length !== filtered.length ? 404 : 200);
		return filtered;
	}
})(request, **response);**
```


xanderkeele-ccpo (approved):
Approved.

